PR #4, opened 2023-06-09
Revive model index

javagl:
The `glTF-Sample-Models` repo contained a [`model-index.json`](https://github.com/KhronosGroup/glTF-Sample-Models/blob/4ca06672ce15d6a27bfb5cf14459bc52fd9044d1/2.0/model-index.json) file that contained information about all available models and variants. This index file was generated with the [`generate-index.py`](https://github.com/KhronosGroup/glTF-Sample-Models/blob/4ca06672ce15d6a27bfb5cf14459bc52fd9044d1/generate-index.py) script.

I have seen the [`allModels.json`](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/5ce2019eb14c553d5a2adbff144250c357a29f23/allModels.json) that also lists all models, but it does not list the variants, file names, and screenshot names.

I think that the `model-index.json` should still exist somewhere. Otherwise, all the tools that rely on this file would be broken. 

Some options:

- Add the missing information about the variants in the `allModels.json`. 
  - This would still be a brekaing change for toolchains that already rely on the `model-index.json`
- Add the `tags` as they are currently contained in the `allModels.json` to the `model-index.json`. This would be backward-compatible: Tools that already read the `model-index.json`  would just ignore the `tags`. 

Further options (depending on the choice above): 

- Extend the current PHP code to generate the information from the `model-index.json`, including the `tags`
- Use the original `generate-index.py` to generate the `model-index.json`, and extend _that_ to include the `tags` there

We could have both, the `allModels.json` and the `model-index.json`, but having only one source of truth would certainly be easier to maintain. (And in any case: This source of truth could be updated automatically with GitHub actions after PullRequests)

### DIFF
--- a/Models/model-index.json
+++ b/Models/model-index.json
@@ -1,15 +1,5 @@
 [
   {
-    "name": "2CylinderEngine",
-    "screenshot": "screenshot/screenshot.png",
-    "variants": {
-      "glTF": "2CylinderEngine.gltf",
-      "glTF-Binary": "2CylinderEngine.glb",
-      "glTF-Draco": "2CylinderEngine.gltf",
-      "glTF-Embedded": "2CylinderEngine.gltf"
-    }
-  },
-  {
     "name": "ABeautifulGame",
     "screenshot": "screenshot/screenshot.jpg",
     "variants": {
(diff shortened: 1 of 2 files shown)


DrX3D:
I am inclined to extend `generate-index.py` to handle the tags for the reasons you list and only create one "index" file. 

The location (.../2.0/) from the old repo does not exist. I am thinking the `model-index.json` file is best in the `Models` directory as it only refers to models. I am also thinking of adding all README collections that are now at the root level to the Models directory.

DrX3D:
The latest PR now updated `model-index.json` and removed `allModels.json`. 

@javagl : Please verify that the repo correctly reflects the intent of this PR.

javagl:
I have not looked at the details of the code that _generates_ the `model-index.json` now, but the current output looks like it indeed makes this PR obsolete.

PR #5, opened 2023-06-09
Updated submission info

javagl:
The [`SubmittingModels.md`](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/5bad5aaa0bbb5d0f9cdc934e626f27d0df1e79b8/SubmittingModels.md) may have to be reviewed. 

Things that are fixed in the current state of this PR:

- Removed an inlined image that does not exist
- Fix broken link to `#converters`. Minor rewording of the variants. Added requirement that models pass validation (!)
- Started updating the description of the `metadata.json`

Here, I stumbled: The description of the JSON there is a bit incomplete, and I'm not sure how to complete this: 

-  What is `version: 2`? (I wondered whether this referred to glTF itself...). 
- Which are valid `tags`? (Yeah, 🐘 -> 🏠 ) 
- Suggestions:
  - `name`: The name of the subdirectory that the model is contained in
  - `path`: The path of the model subdirectory, relative to the repository root
  - `screenshot`: The path to the screenshot, _relative_ to the model subdirectory
  - `summary`: A short description of what the model is or demonstrates
  - `createReadme`: Should always be `true`

The `legal` part also does not seem to list everything that can be found in the actual JSON files. One example is
```
            "license": "CC-BY 4.0",
            "licenseUrl": "https://creativecommons.org/licenses/by/4.0/legalcode",
            "artist": "Cesium",
            "year": "2017",
            "owner": "Cesium",
            "what": "Everything",
            "text": "CC BY 4.0 International",
            "spdx": "CC-BY-4.0",
            "icon": "https://licensebuttons.net/l/by/3.0/88x31.png"
```
That looks like a lot of redundant redundancy, which is pretty redundant. 
- What is the difference between `license` and `spdx`? If `license` is only intended for non-spdx licenses, then I wonder whether this could be merged into `license`, with the values either being one of the spdx-identifiers, or `"Custom"`.
- Is `text` really only for the link text, and does it make a difference of whether the link text is [CC-BY 4.0](http://example.com) or [CC-BY 4.0 International](http://example.com) ?
- That `icon` link... hm. Is this _necessary_ and _useful_? Even if the answer is 'yes': Is "hotlinking" to a random PNG file the right thing to do here...?
- There should be details about the artist/owner. For example, at least I guess: "The names of artists, separated by comma,..."

And... most importantly: What exactly do _submitters_ have to do here? Which of these fields are (or can be) auto-filled: I've seen that link to the `CreateJson.html` (which of course does not work on GitHub...), but is this supposed to fill out most of this information based on less input? Users shouldn't have to figure out the `licenseUrl` on their own, and it would be hard to make sure that the information in the given form is _consistent_ ...

### DIFF
--- a/SubmittingModels.md
+++ b/SubmittingModels.md
@@ -1,6 +1,3 @@
-<p align="center">
-<img src="2.0/glTF_RGB_June16.svg" height="100">
-</p>
 
 # Submitting New Models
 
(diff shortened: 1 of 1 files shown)


DrX3D:
Answering some of the questions.

- This is version 2 of the metadata.json file. It has nothing to do with the version of glTF.
- Technically all tags are valid; however, only a few a recognized by the system when creating the category READMEs. 
- The JSON contains a number of optional fields. These files are filled in when the during various options, such as building the license information (reuse). A lot of the items related to legal are optional from the user's standpoint.
  - `license` is the official name of the license. There are several variants of the CC license that map to the same legal license.
  - `spdx` is the reuse name of the license.
  - `text` is the standardized human-readable version. 'spdx' is almost always determined from 'text'. If a non-standard license is used, special considerations and handling need to be performed.
  - `icon` is really only used for CC licenses. It might need to be enforced.
  - These optional fields could be calculated every time; however, in early development I found it easier to include them in the metadata. The repo build program can automatically update the metadata.json file to include all optional fields.
- The `CreateJson.html` file creates a valid metadata.json file. The creator fills in fields and the populates and displays the JSON structure. That s/w is in beta ATM.

I am not going to claim everything is the correct way here. There is likely a better way to do this, and I am completely open to changing this.

PR #7, opened 2023-07-06
Update to Review Policy

DrX3D:
Multiple updates included in this PR including:

- Updated new and revised model approval requirements.
- Incorporated @javagl changes from PR#5
- Moved model index files to the Models directory

Discussion of missing items in PR#5 has not yet been addressed and must be in the near future.


javagl:
Regarding the last commit, just to keep track of this: The changes from #5 that are not included in the updated state of `SubmittingModels.md` are

- The link in `There is an [experimental HTML application](../util/CreateJson.html)` should start with a _single_ dot `.` (otherwise, it leads to a 404 - but that may not be critical now, as long as this application is not "Live" on a website anyhow...)
- In the sentence `The licensed assigned by the owner`, it should probably be `license` (without `d`)
- The list of license shorthands at the bottom has to be reviewed.

The last point may not have addressed here, but should be sorted out either in an issue or a dedicated PR: We should be _really, really_ careful to make sure that there is a place where the _real_ [SPDX license identifier](https://spdx.org/licenses/) can reliably be found, in a way that is machine-processable and unambiguous

(As in `if (spdxLicenses.includes(thatValue)) { weKnowWhatToDoWithThat(); }`...)

And in these identifiers ...

- `CC0` is called `CC0-1.0`
- `PD` does not exist. I know, the concept of "public domain" is difficult, but `CC0-1.0` is pretty clear from a legal point, and if some pettifogger is not satisfied with that, the model can be provided under a [WTFPL](https://spdx.org/licenses/WTFPL.html) license, dedicatedly issued for the person who does not understand that `CC0-1.0` means exactly that...
- `CC-BY` does not exist. It might be `CC-BY-1.0`. And `CC-BY International 4.0` is called `CC-BY-4.0`

DrX3D:
@javagl : Please take a look at the revised version ([direct link](https://github.com/DrX3D/glTF-Sample-Assets/blob/main/SubmittingModels.md)). I incorporated your changes and redid the descriptions of licenses at the end.

javagl:
Looks fine for me. I guess that I'll close #5 soon then. Even if there are open questions, the PR itself has probably become obsolete.

PR #9, opened 2023-07-07
Remove duplicate MaterialVariantsShoe

javagl:
A copy of the model was contained in the MetalRoughnessSpheres directory


echadwick-wayfair (approved):
Thanks for catching this!

PR #15, opened 2023-07-17
Optimize InterpolationTest

javagl:
Some oddities with the `InterpolationTest` asset have been reported in https://github.com/KhronosGroup/glTF-Sample-Models/issues/390 

And indeed, looking at the model, it contained a lot of "garbage data", and the validator generated 14 Infos and 40 Hints. (Maybe the model was exported with an early version of the Blender Plugin)

I ran the sample through an optimization, basically the same as in https://github.com/KhronosGroup/glTF-Sample-Assets/pull/6#issuecomment-1626126134 , and now, the asset only generates **one** info (about the non-power-of-2-texture)


javagl:
This should become obsolete with https://github.com/KhronosGroup/glTF-Sample-Assets/pull/26

PR #17, opened 2023-07-18
Folder and File Structure added

echadwick-artist:
Added a section with rules for folder structure and file names. Also fixed a few typos elsewhere in the doc.

### DIFF
--- a/SubmittingModels.md
+++ b/SubmittingModels.md
@@ -1,5 +1,5 @@
 <p align="center">
-<img src="https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/glTF_RGB_June16.svg" height="100">
+<img src="Models/glTF_RGB_June16.svg" height="100">
 </p>
 
 # Managing Models in this Repo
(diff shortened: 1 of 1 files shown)


echadwick-wayfair:
There are a few possible points of contention which I would like feedback on.

1. "PNG textures are preferred" ... This seems like it would make sense for the asset in the main /glTF subfolder, to avoid lossy errors in the main asset. But perhaps PNGs should belong in a "source assets" section instead, like used in the old glTF-Sample-Models repo?

2. `/glTF-JPG` subfolder ... Do we want to specify this? Related to the above. If we use this naming convention then the StainedGlassLamp will need to have [its subfolder](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/StainedGlassLamp/glTF-JPG-PNG) renamed to suit.

3. `/glTF-IBL` subfolder ... Do we want to include this? This is only used by one asset so far, https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/EnvironmentTest which uses an extension that is not supported in the glTF Sample Viewer, but is supported in Babylon.js.

bghgary:
> 3\. `/glTF-IBL` subfolder ... Do we want to include this? This is only used by one asset so far, https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/EnvironmentTest which uses an extension that is not supported in the glTF Sample Viewer, but is supported in Babylon.js.

I worked with @MiiBond to add this long ago, but given what we decided with [this PR](https://github.com/KhronosGroup/glTF-Sample-Models/pull/378), maybe we shouldn't include any vendor extension assets (single or multi) in this repo?

DrX3D:
@echadwick-artist , @bghgary : Does anything here require more discussion or approvals. It appears that the questions about folder names, PNG, and screenshot size are unanswered. Would those be better raised as issues?

PR #20, opened 2023-08-04
Automatically check model & create repo files

DrX3D:
Fixed all known issues to automatically check submission and rebuild repo files


DrX3D:
Incorrect request

PR #26, opened 2023-09-06
Improve InterpolationTest

javagl:
"Improve" is a very generic term, so here it is broken down:

- it had a size of 33KB, now it has 18KB
- it contained 10 materials, now it contains 2
- it generated 14 'Infos' and 40 'Hints' (!) when validating, now it generates only 1 'Info' (about the non-power-of-2 image size)
- it had some ... slightly odd, non-looping animations

The last one is the main reason for this update. This was brought up in https://github.com/KhronosGroup/glTF-Sample-Models/issues/391 , with some details described there. The animation data for the linear translations did not loop (i.e. they ended at a different place than where they started), and nearly _all_ values (translations, rotations...) had been ... slightly odd numbers. 

EDIT: The following is now obsolete. Apparently, the bug that caused the model to not be rendered in the sample viewer has been fixed - see https://github.com/KhronosGroup/glTF-Sample-Viewer/issues/495

<strike>Now... merging this should be uncontroversial, **but...** ... the updated model does not render in the [glTF-Sample-Viewer](https://github.khronos.org/glTF-Sample-Viewer-Release/) 😬 </strike>

<strike>I opened https://github.com/KhronosGroup/glTF-Sample-Viewer/issues/495 to track this. Let's see how that plays out...</strike>


javagl:
The `scale` animations in this sample originally interpolated between (0.5, 0.5, 0.5) and (1.0, 1.0, 1.0). The sample now has been updated so that the `scale` animations interpolate between **(0.0, 0.0, 0.0)** and (1.0, 1.0, 1.0). 

The main purpose of this is that according to https://github.com/KhronosGroup/glTF-Sample-Models/issues/321, a scale of (0,0,0) (with 'step' interpolation) is sometimes used to "completely hide" a mesh, and some viewer seemed to not handle this properly.

Issue #27, opened 2023-09-11
Builds are failing


javagl:
As mentioned in https://github.com/KhronosGroup/glTF-Sample-Assets/pull/24#issuecomment-1692400241 (and as it can be seen in the PR list), the builds are currently failing when a new model is added. I zoomed into that, _a little_ bit, but ... am not a GitHub actions expert. So with the usual disclaimers (I have no idea what I'm doing here. LOL), a few comments:

---

Starting with an aside:

The workflows are run _twice_ in the PR that I linked to above. This is probably related to the combination of

- the two events that trigger them, namely `push` and `pull_request`
- @echadwick-artist working directly on the `KhronosGroup` repo, and not on a fork

For example, in https://github.com/KhronosGroup/glTF-Sample-Assets/pull/26 , I'm creating a PR from a fork, and there, the workflows are only run once. Maybe there's a tricky way to avoid running them twice when the PR comes from `KhronosGroup`, but ... it shouldn't be so much an issue. (In fact, there sometimes seem to be difficulties when trying to run such actions based on PRs from a fork, but ... maybe we can make it work both ways). 

---

The actual failing build:

- It **does** work when it is triggered by the `push` event 
- It does **not** work when it is triggered by the `pull_request` event 

Looking at the logs:

- For the `push` case, it ends up saying
  `On branch SpecularSilkPuf` 
  at https://github.com/KhronosGroup/glTF-Sample-Assets/actions/runs/5968588148/job/16192718277?pr=24#step:5:16 

- For the `pull_request` case, it says 
  `HEAD detached at pull/24/merge`
  at https://github.com/KhronosGroup/glTF-Sample-Assets/actions/runs/5968588374/job/16192719075?pr=24#step:5:16

And it cannot commit with a deached head.

Now... why is the head detached here? I don't know. I haven't looked at the implementation of the `actions/checkout@v2` action. From quickly skimming the documentation, it talks about a `ref` property that may determine what exactly is checked out, depending on what _triggered_ the action. This might be related (or similar) to https://github.com/actions/checkout/issues/455.

---

I could reproduce the issue in my "Sandbox" repo (so apparently, ~"something in GitHub" had _changed_ and _caused_ this issue, because it worked smoothly a while ago!). And I played a bit with some some comments from the linked issue, and this change seems to resolve the problem:

https://github.com/javagl/sandbox/commit/297d322b2efd26f3dc1ac477ac047fe79aff5292 

[...]

PR #28, opened 2023-09-11
An attempt to fix the failing builds

javagl:
May fix https://github.com/KhronosGroup/glTF-Sample-Assets/issues/27 , details described there.


javagl:
Giving up here for now. Why should I even care?

javagl:
GitHub actions cannot be used to modify PRs in the way that is required here. We will have to do this manually in the future. 

Go on, prove me wrong 😎

PR #33, opened 2023-10-15
Disable CI. Update README.

javagl:
Fixes https://github.com/KhronosGroup/glTF-Sample-Assets/issues/27


DrX3D:
@javagl : Please check this. I renamed the YML file  (php.yml --> AssetCheck.yml) and changed it to only do check and report. The README change appears to still apply.

javagl:
@DrX3D I'll try to update this PR to include the new check with the `AssetCheck.yml`, and maybe try to create an action that runs when PRs are merged, to update the index files. This would mean that the change in the README could be omitted, and it would automate what is currently done with https://github.com/KhronosGroup/glTF-Sample-Assets/commit/b21d7275efff5ff938e951853a2af2dadcb081c4 ). The latter should not cause any permission issues 🤞 because it's all running in (and only affecting) _this_ repository.

javagl:
Since this PR only deletes the existing file (which should not be done for the final solution), and modifies the README (which should not be necessary in the final solution), I'd rather close this, and try to set up an action that updates the index file when PRs are merged, maybe based on https://stackoverflow.com/questions/60710209/trigger-github-actions-only-when-pr-is-merged ...

PR #40, opened 2023-10-26
Normal maps asset example 

rudybear:
This pull request has 3 different examples how to create normals and normal maps for the assets. Here is a reference that was used: https://bgolus.medium.com/generating-perfect-normal-maps-for-unity-f929e673fc57


echadwick-artist:
Thanks for these assets. 

However it looks like there are some issues with the glTF models:

1. [glTF Validator](https://github.khronos.org/glTF-Validator/) reports missing buffer targets. Ideally these should be fixed. Visual Studio Code with the glTF Tools extension has a quick fix for this
image

2. The glTF models are lacking a good UV layout for baking. The UVs should be consistent texel size, without overlaps, and should have space between UV shells where ever there are hard edges (split vertex normals).

3. The glTF model with the bevels is using all hard edges, but it should use averaged normals. Ideally the vertex normals would be face-aligned for best shading (vertex normals aligned to the normal of the larger neighboring polygon).

4. There are no normal maps included with these models. It would be helpful to include baked maps from Blender, as a demonstration of the process advocated in the tutorial, along with screenshots from the glTF Sample Viewer or another renderer as desired.

Thanks for working on this. Tangent-space normal mapping is a really important aspect of glTF conformance for renderers.

DrX3D:
@rudybear: Please review the 'Asset Quality Workflow' check for a list of errors that need to be corrected before this can be merged. The metadata generator (HTML file in <root>/util) may be of use. Also see [Submitting Models](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/SubmittingModels.md) for instructions, hints, requirements, and recommendations for submitting models.

Also, this repo is not holding the source files for the models. If it is really critical to the proper understanding and/or use of the model, an exception will be made. Please make sure the presence of source files is full documented in the README.

javagl:
@rudybear What is the latest state of this? (I'm a bit surprised that you opened the PR, even though all commits are made by someone else...?)

This was probably supposed to go into the old `....-Models` repo, which may also be the reason why it includes the Blender source files and such (see the previous comment by Leonard). So it looks pretty out-dated. But **iff** the actual glTF model is still supposed to be added, maybe the PR can be updated accordingly.

PR #47, opened 2023-11-11
glTFPotOfCoals added

echadwick-artist:
This asset represents a copper pot using emissive for glowing hot coals, coated with a clearcoat layer on the exterior, and including the glTF and Khronos Group logos.

screenshot_Large


MarkCallow:
I am surprised by the use of .jpgs for the normals and orm (both non-color data). However the rendering at the top of this page looks great.

echadwick-artist:
Thanks Mark! 

I used glTF Compressor to convert the original PNGs (16mb), and the real-time feedback helped a ton with choosing how much compression and downscaling to use. 

This asset is meant for use with the new glTF iOS viewer app, so I wanted to keep the download as small as possible.

We do need to figure out if we want to also include source assets in this repository, like we did with the old glTF Sample Models. That will significantly impact the size of the repo though.

lexaknyazev:
The [Khronos Logo Usage Guidelines](https://www.khronos.org/files/legal/Khronos-Logo-Usage-Guide.pdf) suggest that there should be more clear space around the glTF logo.

DrX3D:
@lexaknyazev : I presume the distance of concern is the space above the logo and below the pot rim. The preferred space is the (vertical size) of "F".  The guidelines allow that the "... clear space requirement may be adjusted to the extent necessary to comply with Visual Integrity requirements". The placement and size of the logo do maintain the visual integrity, so (I believe) it is a matter of artistic choice. 

@echadwick-artist : Would the size & placement of the logo carry the same weight and convey the same effect if it was made to strictly comply with the clear space guideline?

echadwick-artist:
I have resized the glTF logo to match the guidelines, added the DGG logo, and improved various textures.

screenshot_Large
screenshot_Angles

DrX3D:
@echadwick-artist : Please confirm that DGG allows the use of it's logo on 3D models. Most logo usage guidelines do not allow for distortion, which prohibits the use on 3D models. Khronos' guidelines require visual integrity.

echadwick-artist:
I have confirmed, DGG is ok with this use.

Issue #71, opened 2020-10-21
VC model is broken; displays flying boxes.


donmccurdy:
See: https://github.com/KhronosGroup/COLLADA2GLTF/issues/146

https://cx20.github.io/gltf-test/examples/threejs/index.html?model=VC&scale=0.2

cx20:
@DrX3D 

The link is probably broken because the folder name was changed. I think you need to match either the folder name or the file name.

As Is:
|folde rname|file name|
|:------|:-----:|
|VirtualCity/glTF-Binary|VC.glb|
|VirtualCity/glTF-Draco|VC.gltf|
|VirtualCity/glTF-Embedded|VC.gltf|
|VirtualCity/glTF|VC.gltf|

Plan A:
|folde rname|file name|
|:------|:-----:|
|VirtualCity/glTF-Binary|VirtualCity.glb|
|VirtualCity/glTF-Draco|VirtualCity.gltf|
|VirtualCity/glTF-Embedded|VirtualCity.gltf|
|VirtualCity/glTF|VirtualCity.gltf|

Plan B:
|folde rname|file name|
|:------|:-----:|
|VC/glTF-Binary|VC.glb|
|VC/glTF-Draco|VC.gltf|
|VC/glTF-Embedded|VC.gltf|
|VC/glTF|VC.gltf|

javagl:
Despite all the open questions related to this model, the solution should be that of 'Plan A'. 

(Admittedly, I had not looked into https://github.com/KhronosGroup/glTF-Sample-Assets/commit/e5159d0dd6876f220ea180bc7af4b9977a3acfaf by @DrX3D, but maybe just assumed that the file names had been changed as well).

I'll add a TODO on my list and will do that in the next days 🤞

javagl:
The PR is at https://github.com/KhronosGroup/glTF-Sample-Assets/pull/89

PR #87, opened 2023-12-22
Sync 1

DrX3D:
Added several new models that were unique to glTF-Sample-Models. Also updates to the generator program to include a list of model extensions (used and required).

Issue #54, opened 2023-11-20
Set up CI to update the model indices


javagl:
The initial approach for CI in this repository contained a step that was supposed to update the markdown "index" files automatically when a new model is added. When there was a PR, then this CI step tried to update this PR with an update of the index files. Unfortunately, due to some sudden, unannounced, and apparently undocumented change in GitHub actions, this is no longer possible, and it led do https://github.com/KhronosGroup/glTF-Sample-Assets/issues/27. 

Therefore, the CI currently therefore only checks the _validity_ of the models and folder structure (but does not do any modifications on its own). 

It should, however, be possible (and relatively simple) to create a CI step that is triggered when a pull request is _merged_, and updates the index files. It would need to 

- Check out the repository
- Set up PHP
- Run `php util/model.php --no-warn --check --update --process-repo --verbose`
- Commit the modifications
- Push

Considering that this will _only_ run within this repository, by the GitHub bot that _belongs_ to this repository, there should not be any authorization issues with that.

javagl:
This is resolved with https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280 and https://github.com/KhronosGroup/glTF-Sample-Assets/pull/287

PR #89, opened 2023-12-26
Rename VirtualCity files to match asset name

javagl:
The folder name had been changed from `VC` to `VirtualCity`. The individual files still had been called `VC.gltf`/`VC.glb` (as pointed out in https://github.com/KhronosGroup/glTF-Sample-Assets/issues/71#issuecomment-1868449406 ).
 
They are renamed here to `VirtualCity` as well


cx20:
Perhaps it would be better to modify `model-index.json` as well.

https://github.com/KhronosGroup/glTF-Sample-Assets/blob/43380bb399f5c86acb541ee75af4b5905b6926c5/Models/model-index.json#L1024-L1027

javagl:
@cx20 Some aspects still have to be sorted out here. In the future, the model index _should_ be updated automatically, as of https://github.com/KhronosGroup/glTF-Sample-Assets/issues/54 (yes, that's somehow still on my plate as well, and I should give it higher priority). 

(I'm curious whether you'd use the `model-index.json` for `gltf-test` as well - right now, you're maintaining the README manually, is that correct? If you have any thoughts or ideas for what the CI should or shouldn't do, you can mention them in https://github.com/KhronosGroup/glTF-Sample-Assets/issues/54 . It's important to set things up in a way that allows the results to be easily consumed by projects like yours)

DrX3D:
The name change is also done in (not yet merged) PR #87. Since that brings in a several other models and there is a change to the README to include model extensions, I think it would be better to close this without merging. I am just about to merger #87.

DrX3D:
Had a change of mine. #87 will remain open until 4 Jan 2024 to allow people a chance to comment due to the holidays.

Issue #70, opened 2019-02-09
Several models in glTF-Tutorials do not exist in repository of glTF-Sample-Models


cx20:
I remember that registration to this repository was waited because some models had [problems](https://github.com/javagl/gltfTutorialModels/issues/3) with porting from 1.1 to 2.0.
I think that it is desirable to place it in the glTF-Sample-Models repository, since I think that the formal specification of glTF 2.0 is now formulated and the sample in glTF-Tutorials is no problem.
I think the target models are the following three.

- [Simple Material](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_011_SimpleMaterial.md)
- [Simple Texture](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_013_SimpleTexture.md)
- [Simple Skin](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_019_SimpleSkin.md)

@javagl Can you add these models?

PR #90, opened 2023-12-27
Add missing tutorial models

javagl:
Fixes https://github.com/KhronosGroup/glTF-Sample-Assets/issues/70 

Some models that are shown in the glTF tutorial had been missing here. From the ones that are listed in the issue, `SimpleSkin` was already added a while ago. The `SimpleTexture` and `SimpleMeshes` samples are added here.

I also cleaned up some of the README.body.md files of the other tutorial models, and added links to the respective sections of the tutorial.

(I think the `LICENSE.md` files will be created/overwritten during the model index update - but it's all CC0, so there shouldn't be any changes anyhow...)


Models/SimpleMaterial/metadata.json
```diff
@@ -0,0 +1,26 @@
...
+            "icon": "https://licensebuttons.net/p/zero/1.0/88x31.png"
+        }
+    ],
+    "tags": [
+        "core",
+        "testing",
+        "written"
+    ],
+    "screenshot": "screenshot/screenshot.png",
+    "name": "SimpleMaterial",
```
emackey, line 22: I think this name field should have a space. `"Simple Material"`

Models/SimpleTexture/metadata.json
```diff
@@ -0,0 +1,26 @@
...
+            "icon": "https://licensebuttons.net/p/zero/1.0/88x31.png"
+        }
+    ],
+    "tags": [
+        "core",
+        "testing",
+        "written"
+    ],
+    "screenshot": "screenshot/screenshot.png",
+    "name": "SimpleTexture",
```
emackey, line 22: likewise

emackey (approved):

Models/SimpleMaterial/metadata.json
emackey: Actually this repo is very inconsistent about spaces in proper model names.  But it has some, for example:

https://github.com/KhronosGroup/glTF-Sample-Assets/blob/497c9c9f3c254bfce4df64e0fdbac49ce1d8a172/Models/BoxTexturedNonPowerOfTwo/metadata.json#L33

javagl:
@emackey It's actually not soooo inconsistent: Nearly every `name` is just the `ModelName` with spaces to yield `Model Name`.

I just did that change for the newly added and previously existing "simple" (tutorial) models. 

A few short notes that might be related (also @DrX3D ): 

- For some reasons, I thought that `name` was supposed to be _the name_ (as in "the directory- and file name"). So the purpose of the `name` should probably be added at https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/SubmittingModels.md#example-metadata-file . Maybe _roughly_ something like
  > "A short, human-readable name of the model. This will become the link text in the overview page, and the title of the README of the model"

  or so

- The `Box With Spaces` model currently is not displayed properly in the lists:
  Khronos - BoxWithSpaces
  Probably because ... it has spaces 😶 

- I'd suggest renaming the `glTFPotOfCoals` model to `PotOfCoals`. We don't use "Smurf Naming" prefixes

(I'll note these as internal TODOs, but if desired, they can be escalated into issues, to prevent them sinking to the bottom of my TODO list...)

emackey:
> I'd suggest renaming the `glTFPotOfCoals` model to `PotOfCoals`. We don't use "Smurf Naming" prefixes

Agreed.  Also PotOfCoals has a `glTF-Animation` folder, which none of the other animated models have.  Are there GLBs with and without the animation?  Why?  This needs some cleanup too.

PR #91, opened 2024-01-03
Fix relative links in README of Models subdirectory

mehmetoguzderin:
It seems like these files are in the same directory as this README. Although is explicit about being obsolete, it might be more helpful as currently (at least in GitHub Web UI) these take into a file not found error by pointing at the upper directory.  Thank you!

### DIFF
--- a/Models/README.md
+++ b/Models/README.md
@@ -2,10 +2,10 @@
 
 This interface to the model repository is obsolete. You need to use one of the new interfaces from the list below:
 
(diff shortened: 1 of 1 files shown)


cx20:
@mehmetoguzderin Thank you for your advice.
Maybe this will be fixed when PR #87 is merged.

mehmetoguzderin:
@cx20 Thank you very much for your pointer, my bad that I didn't notice the fix in progress! 🙇

PR #92, opened 2024-01-05
Remove AnimatedMorphSphere per Tooling TSG discussion today

emackey:
There are newer models demonstrating morph targets now, and this model doesn't have a recognizable final appearance at the end of the morph, making it impractical to visually inspect the result for accuracy.


javagl:
@DrX3D This doesn't seem to be sooo controversial. If you don't mind, I'd just hit "Merge" here soon.

PR #94, opened 2024-01-05
Add DispersionTest model

emackey:
screenshot-large


DrX3D:
According to [Submitting Models](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/SubmittingModels.md#submitting-new-models) this PR needs 2 approvals. This PR will be merged when that happens.

emackey:
> this PR needs 2 approvals

In general I think that's a poor policy decision for a repo with so few active reviewers.  We would need a vibrant, responsive team of active reviewers to make this work.

MiiBond:
Here's the model rendered in Adobe Substance 3D Stager.
The IBL is different than the one used in Babylon, of couse, but they're both black and white. 
Obviously, they result is very different than with the rasterizer approximation but, then again, refraction is very different too so I don't think we can expect dispersion to match that closely.

DispersionExample

lexaknyazev (approved):

emackey:
Thanks @lexaknyazev!

PR #98, opened 2024-01-29
Adding automatic rebuild of Repo files

DrX3D:
Adds an action that rebuilds the each models Metadata, README, License files and the repo-wide model listing directories.


DrX3D:
@javagl, @emackey  : You might want to take a look at the action in the PR. It is the automatic rebuild on PUSH to `main` that has been discussed for a while.

emackey:
I hope you don't mind if I ask some sanity-check questions:

- What happens if no changes are made by the action, will it attempt to push an empty commit?
- If this action is triggered by a push to main, and itself also pushes to main, does this form an infinite loop?  Or is there a guard against triggering itself already present?
- One of the commented lines mentions the default git push options include `--force`, this sounds terrifying.  Are we sure this is correct?
- I saw [some documentation](https://github.com/marketplace/actions/setup-php-action#github-hosted-runners) saying that `runs-on: ubuntu-latest` already includes PHP pre-installed, but there's still a job called "Set up PHP" here.  Are we sure that's needed?

javagl:
I started quickly looking over this - reading and thoroughly understanding it, by matching it with the documentation of the action at https://github.com/EndBug/add-and-commit (and maybe trying it out in a Sandbox) is something else. 

But a first pass:

---

The settings that are _not_ their defaults are

  `author_name: Author_Name-Test-GitAction`
  `author_email: daly.realism+ActionAuthor@gmail.com`
  `committer_name: Committer_Name-Test-GitAction`
  `committer_email: daly.realism+ActionCommitter@gmail.com`

From skimming over the documentation, it looks like this is not necessary (and you probably don't want all these commits attributed to you). According to https://github.com/EndBug/add-and-commit?tab=readme-ov-file#different-authorcommitter-configurations , it looks like setting the `default_author` (which is currently empty) could fill out these fields with sensible defaults - i.e. setting

`default_author: github_actor`

might just attribute all the changes to "github-actions".

---

In general: The commented-out settings may be helpful during the first pass, but ... when everything is settled, I think that they should be removed, to reduce the clutter.

---

@emackey 

> If this action is triggered by a push to main, and itself also pushes to main, does this form an infinite loop? Or is there a guard against triggering itself already present?

The FAQ says "The commit from the action is not triggering CI!" at https://github.com/EndBug/add-and-commit?tab=readme-ov-file#the-commit-from-the-action-is-not-triggering-ci (with reasons...)

---

> One of the commented lines mentions the default git push options include --force, this sounds terrifying. Are we sure this is correct?

This probably refers to 
```
        # For more info as to why fetching is usually recommended, please see the "Performance on large repos" FAQ. 
        # Default: --tags --force
        # _Default appears correct
        #fetch: false
```
I haven't seen `fetch --force` either, but it seems like this just enforces _overwriting local branches_ (and with a freshly checked out repo in CI, that shouldn't be a problem). 

<sup>(I could point to [the `git fetch --force` documentation](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---force), but ... that whole documentation is the biggest 1f595 ever, and always written in a way that raises more questions than it answers...)</sup>

emackey:
Thanks @javagl!

DRx3D:
Updated action to remove most default settings, comments, and the authoring keys indicated by @javagl. Added all repo-generated files to .gitignore to avoid conflicts.

PR #99, opened 2024-01-29
KhronieAward for testing Dispersion

echadwick-artist:
glTF asset for testing the extension KHR_materials_dispersion.

V-Ray_Render


echadwick-artist:
In the Khronos PBR Technical Sub-Group conference call, we decided to pause this PR. 

This asset uses unorthodox materials for the text behind the glass to circumvent rasterizer limitations. Rasterizers have trouble attempting to render an asset with multiple layers of transmission. 

It would be more physically accurate to use Roughness on the rear of the glass to represent the text, and the surface would have to be doubleSided so the text would be visible from the front, as in reality.

However if the asset was set up that way, most rasterizers would reproduce the asset incorrectly. This would cause a poor user experience for those using the repo for render fidelity testing.

emackey:
This is an awesome model, but per comments above, it's not a good example for glTF beginner users because of the challenges it presents for renderer visual consistency.  I'm going to close this for now, because other people are looking for PRs that need review here, and this needs no further review as-is.  Someday though, I hope this model or a model like this can be merged.

PR #107, opened 2024-02-28
Add dragon with dispersion

emackey:
This is the same dragon model with different settings and a different pattern on the background cloth, to highlight the effects of dispersion on the model.


emackey:
Thanks @echadwick-dgg3d.  I fixed the copyright strings for both `DragonDispersion` and `DragonAttenuation`.   Also I fixed the filename, which had the word "Test" in it even though the parent folder did not.

This is ready.

echadwick-dgg3d (approved):

PR #108, opened 2024-02-28
Fix "Box with Spaces" link issue

### LINKED ISSUE #96: Incomplete Entry for 'Box With Spaces'
'Box With Spaces' does not correctly display in the list of assets. It appears to be solely in the process of creating the asset list markdown code.

emackey:
@DRx3D Can you run the PHP script locally and check if this fixes #96 ?

### DIFF
--- a/Models/Box With Spaces/metadata.json
+++ b/Models/Box With Spaces/metadata.json
@@ -30,7 +30,7 @@
     ],
     "screenshot": "screenshot/screenshot.png",
     "name": "Box With Spaces",
-    "path": "./Models/Box With Spaces",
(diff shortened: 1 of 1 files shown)


emackey:
This PR does not contain the correct fix for the issue.  The PHP code itself needs to be modified to produce `%20` for spaces in Markdown links, such as:

```
[Box With Spaces](./Box%20With%20Spaces/)
```

Note the label text can use real spaces, but the link target cannot contain a space.

It looks like one area of PHP code does contain a string-replace for `%20` called `$pathSafe` or similar, but these safer paths are not used for Markdown link targets.

PR #110, opened 2024-03-01
Added Topologies Without Indices

robertdorn83:
I created a sample file for Triangles, TriangleStrips, TriangleFans, Lines, LineStrips, LineLoops without any indices.


hybridherbst:
@DRx3D was just wondering why this asset was not in the list and realized this PR seemingly never got looked at. Do you see any issues here / something that blocks it?

javagl:
`glTF-MeshOpt` should be `glTF-Meshopt`.

Apart from that, it could be a counterpart of https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/MeshPrimitiveModes (without indices, as the names suggest)

hybridherbst:
Thanks! Do you think it should be a separate asset, or just more geometry on the MeshPrimitiveModes asset?

javagl:
There's the usual trade-off between "fine-grained, specific" models that demonstrate (and allow testing) dedicated features in isolation, and more "comprehensive models" that check several aspects like the different `mode` values. For the dimension of "with vs. without indices", I'd lean towards having two different models. 

As you can see in "Structure" section of the `MeshPrimitiveModes`, I'm usually trying to add _really_ detailed information for these test models to explain what they contain. For the model here, people can only look at the screenshot, and if their rendering looks different, then they'll have to reverse engineer a lot of stuff - e.g. the 'POSITION' values, or which of the displayed primitives is the `LINE_LOOP` and such. Not terribly difficult, but a bit cumbersome, and it could be easy to make it easier...

(A detail: One could argue whether the Draco/Meshopt variants add a lot of value here, but ... 🤷‍♂️ ... OK)

So... I'm not opposed to merging this (after renaming `MeshOpt` to `Meshopt`). Alternatively, I could try to create an example that is visually similar to the `MeshPrimitiveModes`, with similar, detailed information about the structure...

Issue #41, opened 2023-10-27
Changes to Khronos-TextureTransformMultiTest?


bhouston:
I noticed that the render fidelity test framework produces different results for Khronos-textureTransformMultiTest when one uses the model in this repo as compared to the one if the previous glTF-SampleModels repo?

The issue is with clear coat normals.

On the left is Three.js render of glTF-Sample-Models, and the right is using it from glTF-SampleAssets:

![Screenshot 2023-10-25 at 9 37 15 AM](https://github.com/google/model-viewer/assets/588541/8813e7bd-34d3-41ba-88d3-a3da0b608be3)

I also noticed that the Filament engine has similar issues (glTF-Sample-Model on the left, Sample-Aseets on the right):

![Screenshot 2023-10-25 at 9 43 41 AM](https://github.com/google/model-viewer/assets/588541/028cec0d-ceb5-46f5-9d70-ca205178011e)

Other render engines (Babylon.js, gltf-model-viewer) do not have differences.

I figure it must be because of normal reference frame code?

Did someone change the Khronos-TextureTransformMultiTest model in the migration to this repo?

I first noticed the issue here: https://github.com/google/model-viewer/pull/4534#issuecomment-1779299328

emackey:
This is my model, but I'm not the one who changed it between repos.

A quick binary compare of old-vs-new shows that the `TestMap_Normal.png` file has changed, and in particular the old one falsely claimed to be sRGB colorspace in the header, and the new one does not specify that.  This is likely a correction to the model.

Note however, this should make no difference to your renders!  The glTF spec requires that colorspace information from PNG files be ignored, in favor of taking the colorspace as directed by glTF spec itself (which would be linear for a non-color image like this one).  WebGL 1.0 may not be able to abide by this rule, but newer rendering systems should be capable.

See for example:  [TextureLinearInterpolationTest](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/TextureLinearInterpolationTest)

lexaknyazev:
> WebGL 1.0 may not be able to abide by this rule, but newer rendering systems should be capable.

Both WebGL versions support setting this state via `gl.pixelStorei(gl.UNPACK_COLORSPACE_CONVERSION_WEBGL, gl. NONE)` command.

lexaknyazev:
[...]

PR #112, opened 2024-03-15
Correct and clean up PNG normals textures

erich666:
Does not include any JPG or KTX versions of these texture, nor any .GLB files including the textures - those would need to be recompressed. Discussion begun on old Models repo, see https://github.com/KhronosGroup/glTF/issues/2368 for background, etc.

I used my new open-source tool, [NormalTextureProcessor](https://github.com/erich666/NormalTextureProcessor) to check the normals textures in this repo for compliance with the glTF spec.

My notes on the files corrected, and not corrected, are attached. Basically, almost all PNG normals textures have been cleaned up and are in this PR, JPEG normals textures not touched (see notes below), KTX not touched, and GLB files that include normals textures not updated. The JPEGs are lossy, so it's tricky to clean these up. I don't know how to form KTX textures or GLBs, though am happy to be advised and add these to the PR if I must.

Notes attached in notes.txt.
[notes.txt](https://github.com/KhronosGroup/glTF-Sample-Assets/files/14618718/notes.txt)

PR #113, opened 2024-03-19
Update TextureTransformMultiTest normal maps

javagl:
For some reason, the state of the `TextureTransformMultiTest` model here in the glTF-Sample-**Assets** repo does not seem to reflect the latest state that it had in the glTF-Sample-**Models** repo.

This was mentioned in https://github.com/KhronosGroup/glTF-Sample-Assets/issues/41 (and now also noted by @erich666 in https://github.com/KhronosGroup/glTF-Sample-Assets/pull/112#issuecomment-2007968452 ).

The exact reasons are still to be sorted out. For now, this is a PR that replicates the changes from https://github.com/KhronosGroup/glTF-Sample-Models/pull/365 and https://github.com/KhronosGroup/glTF-Sample-Models/pull/367 here - namely, it updates the PNG and the GLB of the model with the fixed normal map.

 Fixes https://github.com/KhronosGroup/glTF-Sample-Assets/issues/41


emackey:
This was fixed in #114.

PR #115, opened 2024-03-26
AnimationPointerUVs

echadwick-artist:
This model tests UV transform animations with [KHR_animation_pointer](https://github.com/KhronosGroup/glTF/pull/2147) for position, scale, and rotation.

screenshot_Large

anisotropyAnimated

[edit: updated screenshots]


Models/AnimationPointerUVs/glTF/AnimationPointerUVs.gltf
```diff
@@ -0,0 +1,13975 @@
...
+            },
+            "path": "pointer"
+          }
+        },
+        {
+          "sampler": 2,
+          "target": {
+            "extensions": {
+              "KHR_animation_pointer": {
+                "pointer": "/materials/14/occlusionTexture/extensions/KHR_texture_transform/rotate"
```
bghgary, line 4924: I think these are all typos? https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/ObjectModel.adoc#5101-interaction-with-khr_texture_transform

```suggestion
                "pointer": "/materials/14/occlusionTexture/extensions/KHR_texture_transform/rotation"
```

Models/AnimationPointerUVs/glTF/AnimationPointerUVs.gltf
```diff
@@ -0,0 +1,13975 @@
...
+            },
+            "path": "pointer"
+          }
+        },
+        {
+          "sampler": 38,
+          "target": {
+            "extensions": {
+              "KHR_animation_pointer": {
+                "pointer": "/materials/23/extensions/KHR_materials_transmission/transmissionTexture/extensions/KHR_texture_transform/rotation"
```
bghgary, line 5320: There are a few of these `transmissionTexture` pointers that point to properties that don't exist.

javagl:
This looks fancy in Babylon.js. (Some look like they are not animated - probably not all properties are supported there yet?).

More out of curiosity: Is the labeling 'U Tiling / V Tiling' intentional, instead of 'U Scale / V Scale'?

bghgary:
We are fixing the issues. There are issues with the asset also. Discussions are happening [here](https://forum.babylonjs.com/t/khr-animation-pointer-with-uv-transforms/49081) if you are curious.

Here is the [latest](https://sandbox.babylonjs.com/?snapshot=refs/pull/14919/merge&assetUrl=https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Assets/8a1af803913e39e21c05df93373f7a782896cbdf/Models/AnimationPointerUVs/glTF/AnimationPointerUVs.gltf).

javagl:
> Should I change the meshes into instances? In 3ds Max it’s possible to assign a separate material to each instance, but I’m not sure if this is possible in glTF?

The geometry data of the spheres is indeed duplicated. So one could probably squeeze out a few bytes with intancing or plainly sharing the `indices/attributes` accessors. But the spheres do not have sooo many triangles, so that's probably negligible compared to the textures. 
(These, in turn, seem to be mostly JPG. I _think_ that a recommendation is to not use JPG for things like normal textures, because the compression may cause very visible artifacts there, **but** ... in this case, the image one of "real tiles", which are a tiny bit "noisy" anyhow - and PNG would probably be twice as large, so JPG seems like a reasonable choice here...)

echadwick-artist:
I have a question about properties sharing UVs. 

The baseColor texture and the alphaMode texture must use the same texcoord and the same uv transforms (KHR_texture_transform), because they are loaded using a single baseColorTexture Property per material (there can’t be two baseColorTextures in a single material).

Similarly, the metallic texture and the roughness texture must use the same texcoord and the same uv transforms, because they are loaded using a single metallicRoughnessTexture property per material (there can’t be two metallicRoughnessTextures in a single material).

However… occlusion uses a separate Property from metallicRoughness, and occlusion may use the same bitmap, or it may use a separate bitmap.

If the occlusionTexture uses the same index as the metallicRoughnessTexture, does this mean they must only use the same texcoord and the same uv transforms?

What about other textures with separate Properties but packed color channels? sheenColorTexture.rgb and 
sheenRoughnessTexture.a for example.

Would the general rule be that if two or more Properties share the same Index, then they must specify the same texcoord and transforms?

lexaknyazev:
> Would the general rule be that if two or more Properties share the same Index, then they must specify the same texcoord and transforms?

There's no rule here. Each texture slot may have its own UV set and transforms.

echadwick-artist:
> There's no rule here. Each texture slot may have its own UV set and transforms.

Thanks Alexey. What do you mean by texture slot? The texture Index? or the texture Property?

donmccurdy:
I would think of it as a material having the following texture "slots" —

- baseColorTexture
- normalTexture
- metallicRoughnessTexture
- occlusionTexture
- ...

These slots ("textureInfo" in spec language) might point to the same or different textures, identified by indices. Each slot may select its UV set and UV transforms independently of the others, even if two slots point to the same texture.

Because there is no glTF concept of an alphaTexture slot at all, it cannot be separated from baseColorTexture. Whereas sheenColorTexture and
sheenRoughnessTexture are separate slots (even if they could optionally point to the same texture) and so their UVs and UV transforms may be separated.

Some engines may be less flexible in supporting all of these permutations, of course. :)

echadwick-artist:
I can't fix the Errors or Warnings, these seem to be caused by lack of support in glTF Validator for the ratified extension KHR_animation_pointer.

echadwick-artist:
I don't understand why the build is failing. The bin and png files are there, and seemingly using the correct case?

2024-11-18 15_25_56-animationPointerUVs gltf - Visual Studio Code

echadwick-artist:
The new glTF file is camelCase but the folder name is PascaleCase. I'll try correcting this.

PR #123, opened 2024-05-19
SheenWoodLeatherSofa

echadwick-artist:
Asset with high quality textures, Sheen and Specular and Webp extensions, and demonstrating the use of three UV methods.

screenshot_Large


rsahlin:
Excellent model @echadwick-artist - great to see one with 3 uv coordinates!
Since this is intended as high quality (hero) asset I see no problem with using PNG textures (or high quality JPEG).
There is definetly value in knowing what the transfer size will be for realistic quality models.
We cannot keep persuing the red-herring that you can get wanted quality with a fixed size budget..... 

Perhaps the default can be PNG - and webp can be in a separate folder just like we have some with Draco compression?

emackey:
Hi @echadwick-artist, beautiful model!

There's an export error here that you might need to work with engineers at DGG to get resolved in software.  You can see the glTF Validator caught it, which is why there's a red "X" on the commit above.

The glTF `textures` list is only allowed to reference PNG and JPG.  So when using WebP, this kind of thing is illegal:

```
  "textures": [
    {
      "sampler": 0,
      "source": 0
    },
```

If the `source` is WebP format, you must use the extension instead:

```
  "textures": [
    {
      "sampler": 0,
      "extensions": {
        "EXT_texture_webp": {
          "source": 0
        }
      }
    },
```

I see you have the extension listed in `extensionsUsed` and `extensionsRequired` already, so that part is good.

This is a technical nitpick, but of course one we would need to have fixed for the sample asset.  Thanks!

lexaknyazev:
> I did use PNG for the big normal map on the wood frame

Have you considered WebP lossless (for all normal maps)? It usually produces smaller file sizes than PNG while keeping the images intact.

PR #124, opened 2024-05-22
Add EXT_meshopt_compression variants of BrainStem and DragonAttenuation

zeux:
This was originally committed in https://github.com/KhronosGroup/glTF-Sample-Models/pull/363

BrainStem was compressed with gltfpack using -vpf -cc (floating-point position quantization + EXT_meshopt_compression)

DragonAttenuation was compressed with gltfpack using -cc (fixed-point position quantization + EXT_meshopt_compression)

There's no particular reason to use -vpf for BrainStem but it just serves as a slightly different encoding example that uses a different set of meshopt filters. Note that in BrainStem most of the data is animation data which ends up being compressed fairly well.

Both models render propertly in Three.JS and Babylon.JS but they do require EXT_meshopt_compression so they currently aren't fully validated (although validator emits no errors).


emackey:
Thanks @zeux.  I'm not sure why these weren't included in the new repo, but it's good to have them back.

PR #126, opened 2024-05-25
Path with spaces

emackey:
@DRx3D There are PHP changes here that should fix the broken Markdown links that had spaces in them.

Also I changed some similar code in a conditional section for `$type == 'Detailed'` but that doesn't appear to run as part of the normal process, so the changes to that section remain untested (but it was certainly broken before).


emackey:
Fixes #96 with a change to PHP code.

emackey:
@DRx3D Anything I can do to help this along?

javagl:
There are some follow-up issues from this not being merged, and things like fixes for https://github.com/KhronosGroup/glTF-Sample-Assets/issues/181 might either be _part_ of this PR, or should be addressed _on top_ of this PR (to avoid merge conflicts). All that to say: It would be good to get this merged rather sooner than later.

PR #127, opened 2024-05-25
PotOfCoals animation pointer

echadwick-artist:
Creating a separate folder for the animated version of the PotOfCoals asset. This asset uses the extension KHR_animation_pointer to animate a heat refraction effect.

PotOfCoalsAnimationPointer


echadwick-artist:
I'm not sure how to resolve these build errors, since they seem to rely on the glTF Validator, but the validator does not support KHR_animation_pointer.

`/extensionsUsed/0: Cannot validate an extension as it is not supported by the validator: 'KHR_animation_pointer'.`

lexaknyazev:
> the validator does not support KHR_animation_pointer

That's not the reason of the failed checks. The root cause is that the old Validator version had a bug that prevented skipping animation pointer stuff.

To roll a new version to Github Action, #135 must be merged first.

PR #128, opened 2024-05-27
DiffuseTransmissionPlant textures changed to JPG/PNG

echadwick-artist:
Changed textures to JPG/PNG instead of WEBP. Stem intersections with obvious shading differences were adjusted so their vertices could be welded together.


emackey:
This follows #122

emackey (approved):

PR #136, opened 2024-07-17
PBR Comparison Samples

echadwick-artist:
Sixteen simple assets created for the new Khronos glTF PBR website, to contrast the omission versus addition of specific PBR features.

2024-07-22_mosaic

[2024-07-22 edit: updated mosiac image]


javagl:
That looks pretty cool. I don't know how "exhaustive" this is (and will have to look at the details) But is this roughy "one 'with' and one 'without' for each (ratified) extension"? 

If so, this could warrant a dedicated "tag" to appear as its own entry in the lists at https://github.com/KhronosGroup/glTF-Sample-Assets?tab=readme-ov-file#lists ...

echadwick-artist:
Basically, yes they represent all the ratified extensions which related to PBR material functions, plus the core glTF 2.0 material functions. 

A few of the samples made more sense to show in triplicate form... alphaMode, baseColor, and Clearcoat. 

For Khronos members, a draft of the website can be viewed here: https://www.khronos.org/members/login/m/api/draft/pbr

echadwick-artist:
> That looks pretty cool. I don't know how "exhaustive" this is (and will have to look at the details) But is this roughy "one 'with' and one 'without' for each (ratified) extension"?
> 
> If so, this could warrant a dedicated "tag" to appear as its own entry in the lists at https://github.com/KhronosGroup/glTF-Sample-Assets?tab=readme-ov-file#lists ...

Currently they are all tagged with #testing ... is this sufficient? If not, what should the new tag be named?

javagl:
I don't have a strong opinion (and maybe others wand to chime in). But the current `testing` already contains many entries. 
(Maybe too many - many the tag is to unspecific...?)

Maybe something like `pbrTesting` or `pbrFeatures` or `pbrComparison` or whatnot - just to narrow it down and make clear that this is a different category than what things like the `AnimatedTriangle` fall in 😁

bghgary:
Are we comfortable with adding a prefix with underscores to the naming of these models? None of the other models have this.

echadwick-artist:
occlusion_before-after
Occlusion changes, before (above) and after (below).

javagl:
@bghgary 

> Are we comfortable with adding a prefix with underscores to the naming of these models? None of the other models have this.

What would be your preference?
I'm a bit on the fence with this. I also don't like the "prefix-with-underscore" either, because it does not match the common naming pattern. One could go so far to say that the goal of grouping these models together is supposed to be accomplished with the "tags", like the `"pbrtest"` tag that has now been introduced. But ... _without_ a prefix, we're left with names like 'Metallic' or 'Ior', which ... aren't ideal. 
This is certainly a new "kind" of contribution here, insofar that it is _one_ set of models with a concept of "belonging together" that is _stronger_ than just the common tag. So I think that _some_ prefix can be justified (maybe just `Compare`, _without_ underscore ...?)

---

On a completely unrelated note: We might have a little problem here @echadwick-artist : It looks like the addition of these models was already done with https://github.com/KhronosGroup/glTF-Sample-Assets/commit/921516f00916741f408716901068207638a9ff1e going directly into main 😬 Not sure what is the best way of cleaning this up. In any case, the state of _this PR_ (when it is ready and supposed to be merged) and the state that is currently in `main` should be carefully aligned.

bghgary:
> maybe just Compare, without underscore ...?

This seems like a good option to me. There should also not be dashes (i.e., EmissiveStrength). Base color should be two words (Basecolor --> BaseColor).

echadwick-artist:
> On a completely unrelated note: We might have a little problem here @echadwick-artist : It looks like the addition of these models was already done with [921516f](https://github.com/KhronosGroup/glTF-Sample-Assets/commit/921516f00916741f408716901068207638a9ff1e) going directly into main 😬 Not sure what is the best way of cleaning this up. In any case, the state of _this PR_ (when it is ready and supposed to be merged) and the state that is currently in `main` should be carefully aligned.

Oops!  I’ll see about rectifying that. Thanks for all the feedback, this is great.

echadwick-artist:
OK I reverted the earlier mistake push into the Main branch. Hopefully this branch is good to go.

PR #140, opened 2024-08-01
Add test of animated base color

emackey:
This model tests the `KHR_animation_pointer` extension's effect on the `baseColorFactor` of a material. When animated, the top cube should rotate above other colored cubes, and change its own color to match the cubes below.

screenshot


emackey:
@lexaknyazev is there someone from Interactivity who would be willing to review this?

It's intentionally simpler than other animation pointer samples.  I suppose it's intended to be the "hello world" of animation pointer, just change the base color and see if anything works.  I believe this is useful to first-time implementers, but I'm curious what others think about it.

Models/AnimatedColorsCube/glTF/AnimatedColorsCube.gltf
```diff
@@ -0,0 +1,441 @@
...
+					"input":12,
+					"interpolation":"LINEAR",
+					"output":13
+				}
+			]
+		}
+	],
+	"materials":[
+		{
+			"doubleSided":true,
```
bghgary, line 106: Why doubleSided?

bghgary (approved):
Small comment. Otherwise, LGTM. I also tested in Babylon.js.

Models/AnimatedColorsCube/README.body.md
```diff
@@ -0,0 +1,11 @@
+## Screenshot
+
+![screenshot](screenshot/screenshot-large.gif)
+
+## Description
+
+This model tests the [`KHR_animation_pointer`](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_animation_pointer) extension's effect on the `baseColorFactor` of a material.  When animated, the top cube should rotate above other colored cubes, and change its own color to match the cubes below.
```
lexaknyazev, line 7: Please add more asset details.

Specifically, the animation features two time inputs: movement is defined for the `[0, 3]` range while colors are defined for the `[0, 2.5]` range. Assuming that it's intentional, this setup additionally checks that the last color keyframe remains active until the movement track is restarted.

Models/AnimatedColorsCube/glTF/AnimatedColorsCube.gltf
emackey: Not really intentional, it's just Blender's default.  Unlike most of my samples, this one actually came out of Blender directly, with no hand-edits.

bghgary: It seems odd to default to something that is likely to run worse. Is it easy to change to not double sided?

emackey: I was able to update this in Blender and re-export cleanly (no hand-edits).

Models/AnimatedColorsCube/README.body.md
emackey: README updated.

emackey:
Pulled out from a now-resolved comment:

> It seems odd to default to something that is likely to run worse. 

This is off-topic here, but it's worth understanding.  Blender chooses its default materials based on what's best for its own purposes, which may not place real-time interactive rendering as the top priority.  The glTF exporter can't tell if the user intentionally chose to have backface culling turned off, or if the user just didn't bother to visit the correct property panel where the choice is offered.

In any case, I believe it's correct for the Khronos exporter to honor the setting in Blender, regardless if it was intentional, and without any guessing as to whether it was altered by the user.  This way, the exported result is a closer match to what's in the Blender preview window.

For these sample assets though, I agree we should be more careful to intentionally choose the better option for glTF.

lexaknyazev (approved):

PR #141, opened 2024-08-03
Update artist name

javagl:
My name was sometimes my full name, and sometimes my GitHub name. This just changes the metadata JSON to be consistent.

I'm not entirely sure where this commit from the GitHub action comes from, and why it changes the `model-index.json`. But this happened automatically, apparently. It only changes the _order_ of certain entries in the JSON, so I assume that this is just some randomness in the order in "dictionaties" in the PHP script...

### DIFF
--- a/Models/AnimatedTriangle/metadata.json
+++ b/Models/AnimatedTriangle/metadata.json
@@ -4,7 +4,7 @@
         {
             "license": "CC0",
             "licenseUrl": "https://creativecommons.org/publicdomain/zero/1.0/legalcode",
-            "artist": "javagl",
+            "artist": "Marco Hutter (https://github.com/javagl/)",
             "year": "2017",
             "owner": "Public",
             "what": "Everything",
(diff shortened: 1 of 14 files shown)


emackey:
@javagl The model-index.json changes order because it's not sorted.  I've added a sort command on line 785 here: https://github.com/KhronosGroup/glTF-Sample-Assets/pull/126/files#diff-09506a581239de149cd51038a4ec19738712ea024006cb0d91a6e070f3246d63R785

It would be great to get #126 merged sometime, as this is an ongoing problem.

emackey (approved):

emackey:
See also: d503b028f3aa835e

javagl:
Thanks for merging and the hints. 

The output from https://github.com/KhronosGroup/glTF-Sample-Assets/pull/126 looks like it fixed the issue. On the one hand, everybody could just hit 'Merge' there at any time (and in doubt, roll back). But when it touches the PHP code, I'd hestiate to do that without revieving this, and I'd need ... ehrm.. a refresher about PHP to really understand the code changes... 

Maybe it's worth another ping @DRx3D to merge https://github.com/KhronosGroup/glTF-Sample-Assets/pull/126 ?

PR #151, opened 2024-10-11
fix: Typo in Everthing (y missing).

atteneder:
The actual change is in the two metadata.json files. I tried to run `util/run.bash`, but did not commit all changes since they seem to be non-deterministic when using it with a newer reuse version on macOS.

### Problems

For further investigation, here's the three discrepancies I did not commit:

reuse 4.0.3 output in `.reuse/dep5.error`:

```
python3.12/site-packages/reuse/project.py:332: PendingDeprecationWarning: '.reuse/dep5' is deprecated. You are recommended to instead use REUSE.toml. Use `reuse convert-dep5` to convert.
```

`.reuse/dep5.spdx` seems to create a different line ending on macOS.

`Models/model-index.json`: the order of the `variants` dictionary is non deterministic.

hth


emackey:
Thanks @atteneder!

Issue #153, opened 2024-11-11
CompareAnisotropy seems to have conflicting normal and anistropy direction


krohmerNV:
This is about one Material in the CompareAnisotropy model, the `grooved-anisotropy` material.

The material looks like this:
```
{
      "extensions": {
        "KHR_materials_anisotropy": {
          "anisotropyStrength": 1,
          "anisotropyRotation": 1.6
        }
      },
      "name": "grooved-anisotropy",
      "normalTexture": {
        "extensions": {
          "KHR_texture_transform": {
            "rotation": 1.5707963705062866,
            "scale": [
              6,
              2
            ]
          }
        },
        "index": 2,
        "scale": 0.5
      },
      "pbrMetallicRoughness": {
        "baseColorFactor": [
          1,
          1,
          1,
          1
        ],
        "baseColorTexture": {
          "index": 1
        },
        "metallicRoughnessTexture": {
          "index": 3
        }
      }
    },
```
Note, it has a rotation of about 90 degrees on the normal texture.
It has also a rotation on the anisotropy direction.
To easier get the point of question, remove both, which should not do much other than rotating the grooves as well as the anisotropy highlight.

Next look at at the normal texture (img2.jpg). The grooves are aligned with the U-Axis or X-Axis
of the image. This axis is usually aligned with the tangent and based on the KHR_materials_anisotropy spec, also with the anisotropy direction (before applying any rotation).

This would also mean that the highlight should also be stretched along that direction, i.e. along the grooves, but it isn't in the reference image: 

`Note: The direction vector of the anisotropy is the direction in which highlights will be stretched. The direction of the micro-grooves in the material causing the anisotropy will run perpendicular.`

My claim is that the normal texture and the anisotropy highlight do not match in the scene setup.
If the highlight stretching is to be aligned to the U-Axis of the texture frame, I would expect the groves in the normal map to be aligned with the V axis so normals and anisotropy highlights match visually.
When applying a 90 degree rotation to both, normal space and anisotropy, nothing changes. Rotating only one would fix the problem but it points to a bug in the reference renderer.

echadwick-dgg3d:
Thanks for pointing this out, looks like my mistake! I'll see about correcting this.

krohmerNV:
Thank you! Apart from this inconsistency in the content I guess the more concerning part is the reference renderer which presumably ignored one of the transformations.

emackey:
[...]

PR #114, opened 2024-03-20
Badly normalized normals texture fixes

erich666:
Some normals textures are poorly formed, with too-long or too-short normals and other problems. This commit addresses the worst PNGs. It does not fix the JPEG versions a few have (see notes below), KTX files, nor makes new GLB files for the corresponding glTF-Binary directories. See https://github.com/KhronosGroup/glTF/issues/2368 for background, etc.

Normals analyzed and cleaned by my new project https://github.com/erich666/NormalTextureProcessor. My notes are below, full analysis report from the program is attached in analysis.txt.
[analysis.txt](https://github.com/KhronosGroup/glTF-Sample-Assets/files/14671585/analysis.txt)

A better fix for the models with problems, in my opinion, would be for the creators to properly form their normals textures from their original heightfield textures. Barring that, this submission at least attempts to clean up these non-compliant files. The main problems found were normals that were not near 1.0 in length, for various reasons (e.g., for some, the blue - Z - channel was never used at all, being 255 for all texels).

Avocado_normal.png - The blue channel values in file 'Avocado_normal.png' all have the same value: 255. That's usually a sign Z is not saved at all. Bad normals, with the Z values (assuming range -1 to 1) found to be as far off as 53 levels in expected value. Normal lengths found in the file, assuming Z -1 to 1: minimum 1.00002 and maximum 1.28911 (these should be near 1.0).

BarramundiFish_normal.png - Reason: the percentage of unnormalized XYZ vectors (assuming Z -1 to 1) is 18.0848% (758533 texels). Normal lengths found in the file, assuming Z -1 to 1: minimum 1.00002 and maximum 1.41678 (these should be near 1.0).

BoomBoxWithAxes_normal.png - 96.4118% of the XYZ normals are normalized well (4043804 texels), 0.304532% are close (12773 texels), and 3.28367% are pretty bad (137727 texels). Normal lengths found in the file, assuming Z -1 to 1: minimum 1.00002 and maximum 1.71857 (these should be near 1.0).

BoomBox_normal.png - 96.4118% of the XYZ normals are normalized well (4043804 texels), 0.304532% are close (12773 texels), and 3.28367% are pretty bad (137727 texels). Normal lengths found in the file, assuming Z -1 to 1: minimum 1.00002 and maximum 1.71857 (these should be near 1.0).

[...]

PR #154, opened 2024-11-18
Anisotropy fix

echadwick-artist:
Removed KHR_texture_transform rotation for the normal map, rotated the pixels of the normal map, and reset the anisotropyRotation to zero. Adjusted anisotropyStrength so the normal map will be visible when the camera is perpendicular to the surface.

Per the reported issue here: https://github.com/KhronosGroup/glTF-Sample-Assets/issues/153

Screenshots from various glTF renderers:

2024-11-18 10_36_24-three js editor

Note that Dassault Enterprise PBR does not seem to support anisotropyTexture, so the spiral grooves are not rendering the same as the webGL rasterizers.


emackey:
@echadwick-artist Can you pull the latest copy of main and merge it down into this branch?  Thanks!

emackey:
The rotated normal map here doesn't look right to me.  The grooves are most pronounced in the green (vertical) channel, not in the red (horizontal) channel.  The map was presumably correct when these were horizontal grooves with long faces pointing +Y and -Y, but now that the image rotated, the long faces need to point +X and -X instead.

I think the correct answer is to swap the red and green channels of the image, and then invert one of them (maybe the green one after the swap).

Also the normal map apparently is itself not normalized (just guessing from the blue channel), and may need the same type of fix as #114.

echadwick-artist:
Oh gosh, you got me haha. I rotated the bitmap, not the normals! How could I forget this is a bunch of vectors, not a just pixels?

PR #157, opened 2024-11-25
Add URI test assets

hybridherbst:
This PR adds various test assets for URIs after I noticed that the allowed formats in the glTF spec include percentage encoding, JSON-string escaping and regular UTF8 strings.

The test assets currently break when dropped in almost all implementations, including the Khronos Sample Viewer. The glTF Validator is able to validate most of them as correct, but not all of them (URIs traversing into parent directories can't be validated it seems).

I included ZIP files for convenient download for testing.


javagl:
A clearer description of each asset would be nice, including the name and what exacty the asset tests or demonstrates, e.g.

> - `glTF-UriTest-03-RelativeBin`: Contains relative references to the `.bin` file and texture that are contained in subfolders
> ... 


The naming of the files is somewhat inconsistent. We should consider using the same naming conventions as for the other models, _even when_ the actual assets are contained in a subfolder. For example, the `.gltf` file in `glTF-UriTest-03-RelativeBin` is called `RelativeResourcePaths.gltf` - maybe it should be called `glTF-UriTest-03-RelativeBin.gltf` for consistency...?

A consistency nitpick: The other file names. There are `Buffer.bin` and `buffer.bin`, and once a `RelativeTexturePaths.bin`...

An aside: I think that this is the only model so far where the _actual_ models are contained in subfolders. In some way this is similar to the "Compare..." assets that have been added via https://github.com/KhronosGroup/glTF-Sample-Assets/pull/136 , and where one could make a case for "grouping" them. But _here_, having them in one "top-level" directory probably makes sense. ....🤞 hoping that this for itself doesn't cause trouble for the CI Python scripts and generated pages.

While I generally see the point of using non-ASCII characters (to give some sense to the different forms of encoding them in the URIs), I wonder whether the names really have to be ` नमस्ते, दुनिया/texture/ChildFolder 你好，世界.jpg`, or whether something like `example/texture/childFolder ①.jpg` wouldn't be sufficient for that.

javagl:
This has some overlap to https://github.com/KhronosGroup/glTF-Sample-Assets/issues/283#issuecomment-4582979245

PR #158, opened 2024-11-30
Add Knight by Quaternius

emackey:
This adds a test model with a skinned mesh, and non-skinned meshes as children of the joints.


Models/Knight/README.body.md
```diff
@@ -0,0 +1,9 @@
+## Screenshot
+
+![screenshot](screenshot/screenshot-large.jpg)
+
+## Description
+
+This is a version of [Knight by @Quaternius](https://quaternius.com/packs/knightcharacter.html) that has been modified to use child nodes with meshes parented to joints.  The Knight character's body is a skinned mesh, but the helmet and the optional sword are separate meshes.  The helmet is parented to a joint in the head, and the sword (which only appears in some of the available animations) is parented to a bone in the right hand.
+
+If the helmet is missing or appears out of place, it may indicate the client implementation has not processed the parent/child hierarchy correctly including the set of joints in the parent.
```
bghgary, line 9: I didn't realize initially that this model had multiple animations. If the first animation `Run` is chosen, there is no sword. I was confused for a bit. It might be worth mentioning the animations and what they are supposed to do somewhere in this doc. Or maybe reduce the animations to a single one that matches the screenshot and description.

PR #160, opened 2024-12-17
A Beautiful Game: added glTF-Binary-KTX-ETC1S-Draco variant

hybridherbst:
Downloading this asset for testing is currently harder than it should be because there's no GLB version.
This variant uses ETC1S compression for textures and Draco for the meshes, and becomes reasonably small without a strong perceptual loss. For the sake of file size I would not add this asset with UASTC textures. I don't see strong visual regressions with ETC1S even for texture slots where usually UASTC should be used.


echadwick-artist:
This looks great to me, thanks for creating this Felix! 

The bishop mesh somehow has 288 less triangles according to glTF.Report, but this doesn't seem to be a problem.

Issue #78, opened 2019-01-09
Is there a model corresponding to KHR_lights_punctual somewhere?


cx20:
I would like to test the model corresponding to [KHR_lights_punctual](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_lights_punctual). 
I found a model below.

https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_lights_punctual/schema/examples
https://github.com/mrdoob/three.js/tree/dev/examples/models/gltf/Monster/glTF-lights

I do not know who to ask for, but if you do not have a problem with rights, I would like you to donate to the glTF-Sample-Models repository.

lexaknyazev:
Here's yet another stress test: max number of light sources.

javagl:
I had accidentally hit a limit here: For that "disco" (random lights) I originally created 16x16 lights, which caused Three.js to bail out with
> Program Info Log: FRAGMENT shader uniforms count exceeds MAX_FRAGMENT_UNIFORM_VECTORS(1024)

With 12x12 lights, it still works. 

Babylon.js seems to limit the number of ligths to 4 (four!) to begin with, so... even the "serious" test model doesn't look that great...

deltakosh:
@javagl: The limit is just a constant you can change (https://doc.babylonjs.com/features/featuresDeepDive/lights/lights_introduction#limitations)

javagl:
@deltakosh I only used the sandbox at https://sandbox.babylonjs.com/ for that test, but it's good to know that this is configurable on the library level.

deltakosh:
@javagl No worries! We are thinking with @bghgary to change that option when loading a glTF with more than 4 lights.
Also for the fun, here is a PG where I generate 160lights:)
https://playground.babylonjs.com/#7MK6JN

PR #163, opened 2025-01-08
Add a test of light intensity vs. color

emackey:
@lexaknyazev Is this similar to what you had in mind?


emackey:
Follows https://github.com/KhronosGroup/glTF/pull/2215

lexaknyazev:
The spec language means that the specified color works as a "filter" in front of a light source of the specified intensity. So we need two more groups of planes to further assert the expected behavior (they all could be in the same asset for convenience):

> Two coplanar squares next to each other. One white point light is located above the center of the first square; three point lights: red, green, and blue, are co-located above the center of the second plane. All intensities must be the same.

> Two coplanar squares next to each other. One point light with color (1, 1, 1) is located above the center of the first square; another point light with color (0.5, 0.5, 0.5) is located above the center of the second plane. Both intensities are the same.

javagl:
This sounds somewhat similar to what I did for https://github.com/KhronosGroup/glTF-Sample-Assets/issues/78#issuecomment-1974027746 (and yes, shame on my, I never opened that PR...)

If desired, I can allocate a bit of time to (really) create such a model. With the nitpick: Is the part of having _separate_ squares relevant for the actual comparison? (It might be, roughly as in "The squares must look equal", as opposed to "The refections of the light on the surface must look equal"). 


(And an aside: Auto-generating such a scene would also avoid these
`"intensity":1.0000660226034708,`
that make my eye twitch sooo bady 😆 )

lexaknyazev:
> Is the part of having _separate_ squares relevant for the actual comparison?

I think that having distinct gaps between squares would make the differences (or the lack of them) more compelling,

javagl:
@emackey I hope I'm not stepping on your toes here.

This would be unit squares with base color (1,1,1) (ok?)
Point lights are placed at y=0.5 at the respective center above them.
All lights have an intensity of 1.0.
The colors are given (verbally) in the labels. I.e. in the upper right, there are three lights. And "Gray" means (0.5, 0.5, 0.5) here.

LightsPunctual Intensity 2025-01-09


The tweaking for the settings in the viewer may have to be explained in the README. Otherwise, users might see the model with default settings:

LightsPunctual Intensity defaults 2025-01-09


The model is here, just for reference for now:


[LightsPunctual Intensity 2025-01.09.glb.zip](https://github.com/user-attachments/files/18365845/LightsPunctual.Intensity.2025-01.09.glb.zip)

emackey:
@javagl I don't mind which of these models gets merged, whatever works best.

I used (0.8, 0.8, 0.8) as the base color of my test material.  I've heard a general rule-of-thumb for PBR that says one should not typically go higher than this in any channel of the base color, to allow for specular highlights and hotspots to have some headroom without leaning completely on tone mapping.  But clearly this rule-of-thumb is for photorealistic objects, not test scenarios like we have here.  Even so, with light sources present, I thought I would use 0.8 gray for the test surface.

emackey:
screenshot-large

Models/PointLightIntensityTest/glTF/PointLightIntensityTest.gltf
```diff
@@ -0,0 +1,868 @@
...
+			"name":"Label Mat",
+			"pbrMetallicRoughness":{
+				"baseColorTexture":{
+					"index":0
+				},
+				"metallicFactor":0
+			}
+		}
+	],
+	"meshes":[
```
lexaknyazev, line 412: Why does the asset have six distinct frame meshes?

Models/PointLightIntensityTest/README.body.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Description
+
+This model tests various colors of point lights (via `KHR_lights_punctual`) that all share the same intensity (1.0).
+
+It is intended to serve as a test for the following clarification that was added to the `KHR_lights_punctual` extension:
+
+> The `intensity` represents the luminous intensity that the light would emit if it were colored pure white (`[1.0, 1.0, 1.0]`). The `color` property acts as a wavelength-specific multiplier.
+
+In other words, all the lights share the same amount of raw energy, but do not share their perceptual brightness.  The blue lamp in particular should appear visually not as bright as other colors.
```
lexaknyazev, line 13: This statement is incorrect. Consider replacing it with:

> In other words, the specified color works as a "filter" in front of a light source of the specified intensity.

emackey: Done.

Scutua (approved):

Scutua (approved):

emackey:
Suggestion from @lexaknyazev: Add the range property to limit light influence.

emackey:
Note for [Babylon Sandbox](https://sandbox.babylonjs.com/) users: The default maximum number of lights is 4, but this model needs 8.  Gary suggests entering this in the console:

```
BABYLON.Engine.LastCreatedScene.getMaterialByName("Test Surface Material").maxSimultaneousLights = 8
```

bghgary:
> Note for [Babylon Sandbox](https://sandbox.babylonjs.com/) users: The default maximum number of lights is 4, but this model needs 8. Gary suggests entering this in the console:
> 
> ```
> BABYLON.Engine.LastCreatedScene.getMaterialByName("Test Surface Material").maxSimultaneousLights = 8
> ```

We fixed it so that Babylon sandbox will auto update the max lights based on what's in the glTF.
https://sandbox.babylonjs.com/?asset=https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Assets/7fa7b20a20b886738839cae83539b26ae637dec1/Models/PointLightIntensityTest/glTF/PointLightIntensityTest.gltf

emackey:
Update screenshot

Models/PointLightIntensityTest/README.body.md
```diff
@@ -0,0 +1,24 @@
+## Screenshot
+
+![screenshot](screenshot/screenshot-large.jpg)
+
+## Description
+
+This model tests various colors of point lights (via `KHR_lights_punctual`) that all share the same intensity (1.0).
+
```
lexaknyazev, line 8: Please mention here that the exact asset appearance is subject to the renderer's IBL and exposure settings and thus may differ from the screenshot.

emackey: Done.

lexaknyazev (approved):

PR #165, opened 2025-02-10
Removed backfaces, adjusted diffuseTransmission

echadwick-artist:
screenshot_Large

* Leaf backfaces removed, and doubleSided:true used instead, to allow pathtracers to render this asset effectively. (Previously the asset was relying on backface culling to render properly, but this is not supported in most pathtracers, which makes it impossible to check for ground truth.)
* baseColorTexture edited so veins are now red. (to help communicate the idea that the leaves are red inside, and thus would show red-colored diffuse transmission)
* Firefly2 animation changed to counter-clockwise, to make it very different from Firefly1.
* Asset re-exported from 3ds Max, and re-optimized with RapidPipeline.
* Textures packed into single shared UV layout, and ambient occlusion re-baked.
* diffuseTransmissionTexture added, to show transmission on stems and leaf veins.


DRx3D:
Since this is an existing asset, should the README be updated with a description of the changes to the model?

echadwick-artist:
I guess I should add something about doubleSided vs. doubled triangles and how that works better with pathtracers, good suggestion.

UX3D-haertl:
The diffuseTransmissionTexture does not use the alpha channel, which is the relevant channel.



> A texture that defines the percentage of non-specularly reflected  light that is diffusely transmitted through the surface. Stored in the  alpha (A) channel. Will be multiplied by the diffuseTransmissionFactor.

PR #171, opened 2025-03-20
New "Transmission Order Test" model

emackey:
screenshot-large

This came out of a conversation with Tim Lewis.


emackey:
I've updated the model based on feedback from Monday's PBR TSG meeting.  There are now 3 rows, showing opaque alongside the other alpha modes.  The blue color has been lightened a bit, to better show when the symbol is behind it.

The image below was path-traced using Blender 4.4.0 Cycles engine, using HDRI from PolyHaven's [Cloud Layers](https://polyhaven.com/a/cloud_layers).

path-traced screenshot

emackey:
There was no further feedback from the PBR TSG when I showed this on Monday.  So, I think it's good to merge now.

Issue #175, opened 2025-04-30
Glass Hurricane Candle Holder - Logo back faces are not seen through the glass as expected.


tim-lewis:
To further enhance the realism of the model, is it possible to have the back faces of the glTF and Khronos logos be seen through the glass, when it is rotated.

![Image](https://github.com/user-attachments/assets/ff4afa36-32a7-4424-b4ce-6c213fff872e)

echadwick-artist:
I tried a few different approaches, but none of them sufficiently solve the issue.

Left to right:
1. Logos textured to the outside surface of the glass material.
2. Logos textured to both the inside and outside surfaces of the glass material.
3. Logos textured to separate triangles, just above the outside of the glass surface, with an alphaMask/doubleSided material.
4. Logos textured to separate triangles, just above the inside of the glass surface, with an alphaMask/doubleSided material.

Three.js rasterizer:
![Image](https://github.com/user-attachments/assets/e1931c91-3ca1-468c-9c83-9067394a20fb)

Option 1 is basically the same as the existing asset. Only the nearest surface shows the logos.
Option 2 also only shows logos on the nearest surfaces. Though if you look into the vase from above, you can see the logos on the inside, but only as long as there isn't glass in the way to block it.

Three.js rasterizer, from above:
![Image](https://github.com/user-attachments/assets/7101d217-af6f-4ff2-8e7d-7f85aa2bb40d)

Options 3 and 4 show the logos on both sides in a rasterizer, but we get distortions due to how the Transmission extension is rendered in realtime. 

Option 3 shows a doubling effect, because the outside logo is being rendered once as normal, but the Transmission glass is rendering the logo again. This happens because a rasterizer is just rendering the whole scene minus the glass, storing this as a texture, then mapping it back onto the glass, plus factoring in distortions introduced by the Volume thicknessTexture which is attempting to simulate how refraction would bend the light rays.

Option 4 is only rendering the logo once, but it shows strong refraction distortions, because it's technically behind the glass.

Three.js pathtracer:
![Image](https://github.com/user-attachments/assets/87e50df5-a869-4ccd-bbde-03238c280cc6)

Not surprisingly, these all look pretty good in a pathtracer. 

Option 1 has some dark fringing around the logos, maybe due to interactions between the baseColor and the transmissionTexture.

Option 2 has a slight doubling effect from having inside and outside duplicates.

[...]

PR #176, opened 2025-05-01
GlassHurricaneCandleHolder logos doubleSided

echadwick-artist:
This PR is to address issue #175.

A duplicate surface has been placed inside the glass, and the logos have been applied there using the glTF properties alphaMode:Mask and doubleSided, so the logos can be seen on both the inside and outside of the glass. 

The rest of the glass has been divided into two materials, to preserve the bands of rough glass above and below the logos, and to prevent the logo roughness from affecting the glass over the logos. 

The baseColorTexture has been edited to add an alpha channel for the logos.

Before/after, rendered in Three.js within Visual Studio Code:
Visual Studio Code


bghgary:
I think there is an issue with the alpha masked texture.
image
The background of the logos is black. Some of the black will bleed into the logo because the texture is sampled with linear/linear. Maybe add a border around the different colored part of the logos to mitigate?

Apart from that, I think using alpha mask looks worse in some engines because it doesn't anti-alias anymore like the original. The edges of the logos of the new model looks worse than the original.

Also, I'm wondering if this new layering will render correctly in a ray tracer.

echadwick-artist:
Tried another approach, using dedicated logo textures (rather than packed into the glass texture like before). and floating the logos as separate polygons above the glass.

Looks good in a pathtracer like three.js in Realistic mode, but looks terrible in rasterizers.

2025-05-04 17_20_08-three js editor


2025-05-04 17_20_44-three js editor
2025-05-04 17_16_28-glTF Sample Viewer

echadwick-artist:
Reposting my conclusions, and closing this without merging the changes.
--------------------------------

I tried a few different approaches, but none of them sufficiently solve the issue.

Left to right:
1. Logos textured to the outside surface of the glass material.
2. Logos textured to both the inside and outside surfaces of the glass material.
3. Logos textured to separate triangles, just above the outside of the glass surface, with an alphaMask/doubleSided material.
4. Logos textured to separate triangles, just above the inside of the glass surface, with an alphaMask/doubleSided material.

Three.js rasterizer:
Image

Option 1 is basically the same as the existing asset. Only the nearest surface shows the logos.
Option 2 also only shows logos on the nearest surfaces. Though if you look into the vase from above, you can see the logos on the inside, but only as long as there isn't glass in the way to block it.

Three.js rasterizer, from above:
Image

Options 3 and 4 show the logos on both sides in a rasterizer, but we get distortions due to how the Transmission extension is rendered in realtime. 

Option 3 shows a doubling effect, because the outside logo is being rendered once as normal, but the Transmission glass is rendering the logo again. This happens because a rasterizer is just rendering the whole scene minus the glass, storing this as a texture, then mapping it back onto the glass, plus factoring in distortions introduced by the Volume thicknessTexture which is attempting to simulate how refraction would bend the light rays.

Option 4 is only rendering the logo once, but it shows strong refraction distortions, because it's technically behind the glass.

Three.js pathtracer:
Image

Not surprisingly, these all look pretty good in a pathtracer. 

Option 1 has some dark fringing around the logos, maybe due to interactions between the baseColor and the transmissionTexture.

Option 2 has a slight doubling effect from having inside and outside duplicates.

Options 3 and 4 both look good, though 4 has some glass attenuation causing some darker shading. Also a triangular shading artifact from the low-resolution mesh. 

Anyhow, this isn't very solvable in a real-time rendering rasterizer.

Issue #179, opened 2025-07-25
Multiple issues with JPEG textures


lexaknyazev:
_Follow-up to https://github.com/KhronosGroup/glTF/issues/2471._

Some JPEG textures used in this repo violate the glTF 2.0 specification, namely they do not contain a JFIF marker and in some cases also contain Exif blocks. Additionally, some JPEG textures contain Adobe's color transform markers, embedded ICC profiles and/or thumbnails.

<details><summary>Why not having JFIF markers is a problem</summary>
<br>

The JPEG standard does not define any color model or channel semantics. An arbitrary JPEG file may contain up to 255 color channels for a single image. This is partially resolved by additional standards like [ITU-T Rec T.871](https://www.itu.int/rec/T-REC-T.871) (aka JFIF), [ITU-T Rec T.872](https://www.itu.int/rec/T-REC-T.872), and [CIPA DC-008](https://www.cipa.jp/std/documents/download_e.html?CIPA_DC-008-2024-E) (aka Exif).

Unfortunately, the amount of possible proprietary and standardized combinations of different markers is so large that it's not feasible for a glTF implementation to support arbitrary JPEG files. Although some non-standard marker configurations are supported by popular libraries and image-editing apps, it's not reasonable to expect glTF implementations to _knowingly_ support them as well.

The most portable and safe option is to require JFIF markers to restrict the JPEG scope as much as possible.

JPEG files not having JFIF markers are already invalid per the glTF 2.0 spec and will be flagged as errors by the next validator version.
</details>

<details><summary>Why having Exif markers is a problem</summary>
<br>
The Exif standard is officially incompatible with JFIF because both specs require putting their respective markers immediately after SOI. Therefore, a valid JFIF file having an Exif block would be invalid as per the Exif specification. On top of that, there are thousands of various Exif fields that may or may not affect image interpretation. glTF implementations should not need to handle that.
</details>

<details><summary>Why having Adobe's color transform markers is a problem</summary>
<br>
Such markers are generally incompatible with JFIF and they are redundant in all existing assets anyway. Their intended purpose is to better support printing systems.
</details>

<details><summary>Why having embedded ICC profiles is a problem</summary>
<br>
glTF 2.0 explicitly defines that all color textures are treated as having ITU-R 709 color gamut with non-linear sRGB encoding. Any custom ICC profile must be ignored. Opening a JPEG image with a custom color profile in an image editor may yield different results than using the same image as a glTF texture.
</details>

[...]

PR #180, opened 2025-07-25
Replace some JPEG textures with PNG

lexaknyazev:
First step towards fixing #179.


emackey:
I assume these were all tested with a newer draft of the validator?

lexaknyazev:
Yes but that's technically irrelevant - new files are PNGs.

emackey (approved):
Looks good to me.  I'll leave this open for @javagl to take a look too.

javagl (approved):
I looked at the `.gltf` files, and whether the changed textures are reflected with the URL and MIME types, and did not see an error. 

I loaded the GLBs in the sample viewer, and they seem to all be displayed correctly.

I did not check whether all PNGs are "the same" as the corresponding JPGs, but for the ones that I checked, this was the case.

I'm a bit curious about how you created the PNGs from the JPEGs that contain _text_: It looks like these are _the exact same files_, but _without_ the JPEG compression artifacts. I assume that you used some special "denoising" tool/filter for that?

(I'll try to ignore the distress that is caused by the fact that "checkerboard" has 18x18 tiles (and not 2<sup>4</sup>x 2<sup>4</sup>), and that the gray values are 129(!) and 201 instead of 128 and 192 🤓 )

PR #184, opened 2025-07-29
Fix broken path for model tag links in README files.

### LINKED ISSUE #181: Tag links not working
The pages like https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/TriangleWithoutIndices have the "Tags", like 'core', which link to https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models-core.md - this is a 404. There is one `/Models` part missing there. The right link is https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/Models-core.md

emackey:
Fixes #181.

Also removed an obsolete commented-out section.

### DIFF
--- a/util/modelmetadata.php
+++ b/util/modelmetadata.php
@@ -243,16 +243,6 @@ class ModelMetadata
 			$this->metadata[$key] = $value;
 		}
 		
-/*
- *	// Special testing for "weird" directory paths
-		if ($path == './Models/Box With Spaces') {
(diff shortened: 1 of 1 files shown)


javagl (approved):
From only _looking_ at the code, it seems like this might fix it. And assuming that you tested it locally, this looks fine for me. (If I should test it locally, drop me a note)

PR #238, opened 2025-10-23
Losslessly optimize screenshot PNG files using oxipng

aaronfranke:
This PR losslessly optimizes PNG files using `oxipng -o max --zopfli --fix`.

This keeps the quality of the files exactly the same, but reduces the file size. This decreases the download and storage size required for these assets, at the cost of increasing the size of the Git history.


lexaknyazev:
FYI: #239 and #240.

@aaronfranke As of now, it's safe to use `--strip all` on the remaining screenshot files. Are you interested in re-running the optimization but only on `Models/*/screenshot/*.png` paths? If so, this PR could be repurposed for that (with rebase/force-push).

lexaknyazev:
Sorry, I meant updating only the screenshots in this PR. We'll continue replacing/optimizing asset textures one by one, while keeping GLB/glTF variants in sync (the CI will eventually require that).

aaronfranke:
@lexaknyazev Ah, I misunderstood, I've removed all but the screenshot changes.

lexaknyazev:
Looks like something went wrong during rebase and four unrelated JPEG files (previously updated in #241) have been unintentionally changed.

PR #244, opened 2025-11-03
Add test for `KHR_node_visibility`

javagl:
An asset for testing the `KHR_node_visibility` extension.

A first version of this was posted in [a comment in the pull request](https://github.com/KhronosGroup/glTF/pull/2410#issuecomment-2192299602). A version with minor additions was posted in a [later comment](https://github.com/KhronosGroup/glTF/pull/2410#issuecomment-3465258424). It could make sense to add this as a dedicated test asset for the extension here.

The asset uses some nested node structure to display the messages and support status. A summary:

- First row:
  - There is a label that says "The extension is supported"
  - Slightly in front of that, there is a label that says "The extension is not supported"
  - The label in the front has the extension with `visible:false`. So when the extension is supported, this is not shown.

- Second row:
  - There is a label that says "Does not handle defaults properly"
  - Slightly in front of that, there is a label that says "Does handle defaults properly"
  - The label in the front has `visible:undefined`, which should default to `true`. So when the default is handled properly, then this is shown.

- Third row:
  - There are two labels that say "Does not properly handle hierarchy". 
  - These are the leaves of a chain of nodes with visbility flags `false, false, true` and `true, true, false`, respectively. 
  - So they should eventually both not be visible.

(There are some blank rectangles that hide the implementation messages when the extension is not supported at all).

Direct preview to the README body: https://github.com/javagl/glTF-Sample-Assets/blob/add-node-visibility-test/Models/NodeVisibilityTest/README.body.md


javagl:
@bghgary I had seen the 🎉 , but maybe you want to have another look at this to see whether it can be merged. I've tried it out in the Sample Viewer and BabylonJS, and ... that's nearly boring, because they both support the extension. But I've used the BabylonJS "Scene Explorer" for mimicking the forms of "wrong behavior" that are covered via the screenshots in https://github.com/javagl/glTF-Sample-Assets/blob/add-node-visibility-test/Models/NodeVisibilityTest/README.body.md , and hope that this all makes sense now.

Models/NodeVisibilityTest/metadata.json
```diff
@@ -0,0 +1,24 @@
+{
+    "version": 2,
+    "legal": [
+        {
+            "license": "CC0-1.0",
+            "licenseUrl": "https://creativecommons.org/publicdomain/zero/1.0/legalcode",
+            "artist": "Marco Hutter (https://github.com/javagl/)",
+            "year": "2017",
```
bghgary: `year` should be 2024 (first posted) or 2025 (this revision), not 2017.

bghgary (approved):
[Reviewed by Copilot on behalf of @bghgary]

LGTM except for one minor comment.

Models/NodeVisibilityTest/metadata.json
javagl: Greetings to whoever invented "Copy and Paste". This is fixed now.

Thanks for the review.

PR #246, opened 2025-11-14
Trim trailing whitespace and ensure Markdown files are POSIX compliant

aaronfranke:
~~This PR updates the PHP scripts to output POSIX-compliant text files, which means text files that end in one `\n` character.~~ ~~This PR also runs the scripts~~ ~~and manually goes over files the script did not touch (like source `.gltf` files).~~ I also added an `.editorconfig` file to give a hint to IDEs that they should automatically fix this problem if encountered. This PR also trims trailing whitespace.


aaronfranke:
@javagl There are many utilities that expect text files to be POSIX-compliant text files, especially on Mac/Linux systems. For example, `cat` will behave incorrectly in those cases, with the next terminal prompt appearing on the same line as the file's "final line" (in quotes because, by POSIX standard, a line of a text file ends in `\n`, therefore any bytes following the final `\n` are extraneous characters not on any line of the text file per the standard). For example, GitHub itself will display warnings when text files do not end in a `\n` character. For example, C/C++ compilers will complain about this: older compilers had it as an error, newer ones have it as a warning, but still if you compile with warnings treated as errors then it won't compile. This repo isn't C/C++ code, but still, many users including myself have their IDEs configured to automatically fix this problem on save. So, to the point, having POSIX-compliant text files in every repo makes it easier to work with, for me and anyone else who has their IDE configured like me. Otherwise I either need to reconfigure my IDE temporarily, or manually discard changes in Git. Aside from that, I would argue having files consistent is beneficial for its own sake, and consistent with a standard is even better. Formatting is too important to ignore, but not important enough to spend time on, so it's good to have easy-to-follow rules.

As for reviewing, click the "Hide whitespace" check box in the "Files Changed" tab to make reviewing easier:

Screenshot 2025-11-14 043345

javagl (approved):
Summarizing some discussion (also from the Tooling TSG call): 

There had been some concerns about formatting `.gltf` and `.svg` files in a particular way. Every time that a `.gltf` was updated (or a new one was added), the state would "diverge" again from the _current_ state here.

One thing that would make sense and still can be considered:
We would _like_ to have some human-readable formatting for the `.gltf` files. People shouldn't submit `.gltf` files with `{ everything: { in: { one: line }}}` for example. This cannot sensibly be made a _requirement_ for submitting and merging a PR (because we don't have influence on the _tools_ that generate these files). But it could be made part of CI: Whenever a PR is merged, the CI could do some trivial pseudocode of
`write(read("file.gltf"), "file.gltf");`
to apply consistent formatting to the glTF files.

---

That said, _this_ PR was now updated to only affect the markdown files, and _specifically_ the `body.md` files. This is important: These are the files that are contributed _externally_ - in contrast the the actual `README.md` for each model, which is automatically generated, and only inlines the respective `.body.md` file.

So this PR brings all the `body.md` files into a consistent shape. Looking at the changes, it is indeed only 

1. removing unnecessary trailing spaces
2. removing unnecessary newlines
3. adding newlines at the end of the files where they are due

So I think that this can be merged.

---

There had been some inlined comments about the PHP, e.g. at https://github.com/KhronosGroup/glTF-Sample-Assets/pull/246#discussion_r2527369396 . Given that the PHP was changed to TypeScript as of https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280 , I assume that these comments are obsolete. This involves the (somewhat bold) assumption that the new, TypeScript based tools generate MD that does _not_ have the issues that have been fixed here. But this may have to be confirmed. Which leads to the last point:

There should probably be some CI step for (automated) Markdown formatting. I've opened https://github.com/KhronosGroup/glTF-Sample-Assets/issues/290 for tracking this.

javagl:
So I'll just go ahead and merge this. 

@aaronfranke If you want to address https://github.com/KhronosGroup/glTF-Sample-Assets/issues/290 , just drop me a note. Otherwise, I'll try to have a look at this soon (before there are too many additions that "diverge" from the current state)

PR #247, opened 2025-11-14
Fix missing text versions of models

aaronfranke:
This is just the changes to `NodePerformanceTest` and `PlaysetLightTest` from PR #237. They were missing text-based `.gltf` versions of the assets, so the PHP script was producing warnings for these when ran on all files. This PR adds text-based versions to fix these warnings.

Additionally, as requested, I ran `oxipng` on these models (max settings, both deflate and zopfli), and updated the `.glb` versions.


javagl:
The [comment](https://github.com/KhronosGroup/glTF-Sample-Assets/pull/237#issuecomment-3528084144) said

> Could the `NodePerformanceTest` and `PlaysetLightTest` updates be split into separate PRs? 

It did _not_ say

_"Could the `NodePerformanceTest` and `PlaysetLightTest` updates be split into a separate PR?"_ 

This is _not only_ "spec-level nitpicking". There's a reasoning behind that: It would be preferable to have "self-contained changes". This should _at least_ be on the level of commits. One commit "Changed summaries", one commit "Added text version of X", and one commit "Added text version of Y" would be easier to review, easier to handle, and would produce a cleaner history. 

(Whether or not these are individual PRs may then be less important, but I'll let @lexaknyazev decide here...)

javagl:
Don't sweat it. That "self-contained commits" are one goal is still an assumption on my side. That it should be "(two) PRs" instead of "(one) PR" is just my tendency to assume that Alexey is usually _very_ careful with his wording 😀 

There certainly are different approaches for the granularities, and that depend on the nature of the project, and on whether the maintainers and contributors have the same habits or attitudes. In this specific example, the granularity of "one PR (maybe even with just one commit) _per model_" seems to lend itself. Purely subjectively, I'm leaning to the "fine-grained, self-contained" approach (and have only used the `squash` function once or twice in my life until now...). But more generally: It's usually easier to mush many small things together (e.g. squashing 100 commits into one) than identifying what/where/why certain changes have been made in one commit that affects 100 files.

PR #248, opened 2025-11-14
Fix missing DGG license

aaronfranke:
This PR takes out a part of #237, just the fixes for the DGG license.

The Darmstadt Graphics Group GmbH legal mark license was missing validation in the PHP script. This caused the PHP script to produce invalid license text in the `.reuse/dep5` file. This PR adds a case for this license, similar to the existing code for Cesium, Khronos, and UX3D. I also ran the PHP script locally.


javagl:
This looks "OK" for me insofar that it follows the pattern of other licenses. But I'm starting to question that pattern in https://github.com/KhronosGroup/glTF-Sample-Assets/issues/253 (somewhat unrelated to this PR, and probably to be dicussed independently)

PR #250, opened 2025-11-14
Fix missing text version of PlaysetLightTest model

aaronfranke:
This PR adds a text version of the `PlaysetLightTest` model. It was missing text-based `.gltf` versions of the assets, so the PHP script was producing warnings for this when ran on all files. This PR adds a text-based version to fix these warnings. I also updated the light intensities so that the scene is no longer extremely overbright.

Additionally, as requested, I ran `oxipng` (max settings, both deflate and zopfli), and updated the `.glb` version.


lexaknyazev:
This PR opens an interesting question about the `asset.generator` property.

- The original GLB file says: `"generator": "Khronos glTF Blender I/O v4.0.43"`, likely because the asset was actually exported from Blender.
- The new files from this PR say: `"generator": "glTF-Transform v3.10.1"`, likely because that tool was used to expand the GLB and reassemble it.

Should we preserve the original generator value with such maintenance changes or let the latest used tool replace it?

@javagl @emackey @DRx3D what do you think?

javagl:
This question is not ultimately sparked by this PR. For my glTF processing libraries, I considered something like 
`if (generator == null) generator = "JglTF"; else generator += ", post-processed with JglTF";`
to keep some sort of "provenance" of where the model came from.

This alludes to what could be the relevant question here, which is more broadly about the "purpose" of the `generator` field. 

My understanding is that it is mainly for tracking the tool (and maybe the version of the tool) to figure out who to blame when something goes wrong. And while it's certainly possible to write some manual tweaks e.g. for glTF-Transform to _keep_ the old `generator`, that's certainly not the default behavior (for any tool, probably - just _keeping_ what is there would defeat the aforementioned purpose after all...). 

The question is somewhat related to the question that I brought up earlier (e.g. when you updated all the images), namely whether we could or should rely on a _single_ tool generating all the GLBs. For example, I think that glTF-Transform did (and maybe still does) write _interleaved_ attribute buffers _by default(!)_. And we probably don't want all models to only use interleaved buffers. (The same applies to any "default behavior" that is exposed by some tool).

That said: The GLB was (now) not generated with "Khronos glTF Blender I/O v4.0.43", that's a fact. So it should be "glTF-Transform v3.10.1". Depending on how important we think the "history" is, we could check how difficult it is to make that
"Generated with Khronos glTF Blender I/O v4.0.43, post-processed with glTF-Transform v3.10.1"
or so...

lexaknyazev:
The tool built into the VS Code glTF plugin (right-click on a GLB file there, select "Import from GLB") can be used for that. That tool generally preserves everything except buffer views containing images and would name the image files like `PlaysetLightTest_img0.png`. Ideally, we should use more reasonable names but that behavior is at least self-consistent as the file names simply correspond to the image object indices. Also that tool indents generated JSON with 2 spaces instead of 4.

aaronfranke:
Done using the standalone tool. This preserved the original generator information.

lexaknyazev:
@aaronfranke Why were the punctual light intensities changed?

Issue #253, opened 2025-11-14
Link texts and descriptions generated from the metadata


javagl:
I noticed this in the context of https://github.com/KhronosGroup/glTF-Sample-Assets/pull/248 , but it is unrelated to that PR. I started it as a comment, but became long enough to be summarized in an issue. (I didn't thoroughly review everything that was previously added to this repo) 

Mainly @DRx3D : 

There is an array of predefined licenses in the `modelmetadata.php`. The elements have properties `icon`, `link`, `text` and `spdx`. For example [the `text` for "CC-BY-4.0 International"](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/c5c40347a11a849cb58a332e2c7464964bf989f0/util/modelmetadata.php#L162) is

"CC-BY 4.0 International".

This appears as the link text in the ["Legal" section of the generated README.md](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/c5c40347a11a849cb58a332e2c7464964bf989f0/Models/ChronographWatch/README.md#legal).

For Khronos (and some other licenses), that text says

"Khronos Trademark or Logo"

which looks a bit odd as a link text. 

There also are some other glitches. For example, in that linked README.md, there is that bullet point

> * Non-copyrightable logo for Khronos logo

which apparently generated from [the metadata](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/c5c40347a11a849cb58a332e2c7464964bf989f0/Models/ChronographWatch/metadata.json#L18) from a pattern like

> * `artist` for `what`

A broader question could be about the role of the `LicenseRef-LegalMark-Khronos`, which is apparently used in cases where something "non-copyrightable" is involved. This is also used as the `spdx`, but it is not one of the SPDX identifiers listed at https://spdx.org/licenses/ , which may raise questions.

DRx3D:
I an not quite sure where or what the specific question is, so I'm going with discussion.

Trademarks and copyrights are owned IP. Creative material is (usually) copyrighted. Trademarks are usually not copyrighted. These confer differing types and levels of protection. The Khronos trademarks glTF (R) cannot be copyrighted because there is no creative content. It can (and is) trademarked. The glTF logo ("glTF" text with a oval swish) is a trademark. There is some ambiguity in Khronos exactly how that should be listed. 

[...]

Issue #252, opened 2025-11-14
Typos in scripts


lexaknyazev:
https://github.com/KhronosGroup/glTF-Sample-Assets/blob/f4207789b96f757b462c1b1f1b87da535e359610/util/model.php#L218

https://github.com/KhronosGroup/glTF-Sample-Assets/blob/f4207789b96f757b462c1b1f1b87da535e359610/util/model.php#L226-L227

https://github.com/KhronosGroup/glTF-Sample-Assets/blob/f4207789b96f757b462c1b1f1b87da535e359610/util/model.php#L498

https://github.com/KhronosGroup/glTF-Sample-Assets/blob/f4207789b96f757b462c1b1f1b87da535e359610/util/modelmetadata.php#L236

https://github.com/KhronosGroup/glTF-Sample-Assets/blob/f4207789b96f757b462c1b1f1b87da535e359610/util/modelmetadata.php#L483

https://github.com/KhronosGroup/glTF-Sample-Assets/blob/f4207789b96f757b462c1b1f1b87da535e359610/util/modelmetadata.php#L543

Note that after fixing the scripts `LICENSE.md` files will be regenerated.

javagl:
And licenses...

https://github.com/KhronosGroup/glTF-Sample-Assets/blob/c5c40347a11a849cb58a332e2c7464964bf989f0/LICENSES/LicenseRef-LegalMark-Khronos.txt#L3

"incoirporated" -> "incorporated"

PR #254, opened 2025-11-16
Cleanups for utilities

javagl:
So this is somewhat experimental (maybe not to be merged). It addresses https://github.com/KhronosGroup/glTF-Sample-Assets/issues/252 and (to some extent) https://github.com/KhronosGroup/glTF-Sample-Assets/issues/253 . 

The first commit is _only_ formatting. (I just picked the first PHP formatter that I found. I think that automatic formatting can be important, but don't give a ... care about _what_ the _exact_ formatting is).

The next commits are fixing the typos from https://github.com/KhronosGroup/glTF-Sample-Assets/issues/252

Two commits are for some minor formatting of the license texts, and https://github.com/KhronosGroup/glTF-Sample-Assets/commit/ee82cb73fad0cbc790809758e30468fd64364459 adds a text to the "Stanford" license file, which until now, only contained
"See the page on ...tbd."
**But...** I think that this file is not actually used. Maybe it could simply be removed (similarly for some other files in the LICENSES folder)

In the last commit, I tried to fix the license links that didn't lead anywhere, but _should_ have led to the LICENSES folder. There, I stumbled a bit. I'll have to read the PHP more thoroughly (and in the course of that, I'd probably add **comments!!!**). It looks like there are two ways about how this can be run, and one of them inserts some `licenseUrl` into the `metadata.json`, unless that `licenseUrl` was already there. Even without a deeper understanding of what the PHP is doing _exactly_: I think that it should **NOT** modify the `metadata.json`, ever. This is supposed to be pure _input_. It can be checked and validated and the CI can fail when it's not complete and whatnot. But it should **NOT** be modified by CI.


javagl:
Now I wonder whether there's an easy way to ~"trigger a rebuild, and see what the repo would look like after these changes". Have to spend more time for all that...

emackey:
This thread has several different wish lists of things that would be great to have eventually.  But in the interest of small, incremental changes that keep our velocity (just barely) above zero, I think I'll merge this as-is for now.  Once we see the results of this we can reassess and add new issues or make further changes in a new PR as needed, to keep the momentum going.

javagl:
It's not entirely clear what this refers to - the https://github.com/KhronosGroup/glTF-Sample-Assets/blob/a02ccb86d94e546a58ebe9e3a825ae93ce509c6f/Models/Models-showcase.md that has been rebuilt here looks fine for me...?

EDIT: Ah, this refers to these "LicenseRef" links? Yeah... some `../` here and there... 😬

PR #257, opened 2025-12-15
Add MeshoptCubeTest model

zeux:
Specification: https://github.com/KhronosGroup/glTF/pull/2517

This model is generated by the script from [the following Gist repository](https://gist.github.com/zeux/d26340c53dd70d19ae18045e79d065df) using `node generate.js` for glTF-Meshopt version, which has no fallback data and *requires* the extension, and `node generate.js true` for glTF version, which has a fallback binary buffer and can be loaded by any viewer, but viewers that support KHR_meshopt_compression will not need to read the fallback buffer.

Screenshot From 2025-12-15 14-58-37

The asset contains a grid of cubes; each column is using the same source data and vertex layout (in order: interleaved with quantized 8-bit normals/colors; separate streams with quantized 8-bit normals/colors; separate streams with quantized 16-bit normals/colors; and the last column is using uncompressed geometry with a 16-bit quantized quaternion animation track). Each row is using a different set of modes and filters with KHR_meshopt_compression extension (in order: no compression; v0 vertex compression and INDICES index compression; v0 vertex compression and TRIANGLES index compression; v0 vertex compression with exponential, octahedral, color and quaternion filters; v1 vertex compression with the same set of filters).

The leftmost column is using interleaved data and as such is not using filters even when the other elements of the same row do use it, because filters only work for a specific set of byte strides.

All colored cubes are expected to look the same. All animated cubes in the rightmost column are expected to look the same (no colors on this column) and rotate in sync.

To load glTF-Meshopt variant you need a viewer that supports KHR_meshopt_compression; it loads and displays correctly in three.js after https://github.com/mrdoob/three.js/pull/32163 (assuming MeshoptDecoder object passed to the GLTFLoader supports KHR functionality, which is true as of meshoptimizer 0.25).


Models/MeshoptCubeTest/metadata.json
```diff
@@ -0,0 +1,25 @@
+{
+    "version": 2,
+    "legal": [
+        {
+            "license": "CC0",
+            "licenseUrl": "https://creativecommons.org/publicdomain/zero/1.0/legalcode",
+            "artist": "Arseny Kapoulkine",
+            "year": "2025",
+	    "owner": "Public",
```
lexaknyazev: nit: mixed tabs/spaces formatting on this line; please use spaces

lexaknyazev:
Please add `Models/MeshoptCubeTest/README.body.md` with the asset description; see other similarly named files in this repo for the template.

The PNG files with label textures use 16-bit precision and contain extra metadata. The screenshot image has unneeded alpha channel and extra metadata. Please optimize them by running the following command and force-push the image updates so that only the optimized files reach the main branch.
```
$ oxipng --zopfli -o max --strip all --scale16 glTF/*.png glTF-Meshopt/*.png screenshot/screenshot.png
```

javagl:
I had seen this PR, and quickly scrolled over the linked gist, but not yet done a real "review". 

The first part of the review would be to check whether there is important information in the Gist that is _not_ part of the README (I've seen that there _is_ a lot of information in the README, but haven't cross-checked it). Depending on the intended depth of the review, I may have to allocate a bit more time for that.

From a first click on the "Files Changed": The files are called `scene.gltf`, but should be called `MeshoptCubeTest.gltf`.

PR #258, opened 2025-12-15
Update BrainStem and DragonAttenuation to KHR_meshopt_compression

zeux:
Specification: https://github.com/KhronosGroup/glTF/pull/2517

The updates are performed by converting the source models with gltfpack 1.0, using -cz for DragonAttenuation and -cz -vpf for BrainStem (consistent with the original conversion settings), and post-processing the glTF files using `jq` as a pretty-printer.

The resulting files render the same, but require the viewer to support the KHR variant of the extension. While this is typically very easy to do (see https://github.com/mrdoob/three.js/pull/32163 for example), as of this PR no viewer supports it yet, so we may want to wait to merge this until a couple viewers get the KHR support.

The updated files here are not using fallback buffers so the extension is required; see MeshoptCubeTest model for a more complex and varied set of tests that exercise the JSON structure more broadly, this PR is focused on updating the more real-world models so that we have both a synthetic test version and a more realistic one.


javagl:
Having these models _somewhere_ is good, and I see the reasoning from the context of the discussion at https://github.com/KhronosGroup/glTF/pull/2517#issuecomment-3657864685 (~"there must be sample models for the ratification")

**But...** (probably mainly @lexaknyazev , but also the other "usual suspects") : 

Should the original EXT meshopt versions be retained? 
(I.e. should the new ones be added in a _new_ directory?)

I'm leaning towards _keeping_ the original ones, and adding the new ones. This mainly rooted in ... certain aspects that may not have been thought through back when the structures here have been established: 

- The directory name `glTF-Meshopt` gave a very prominent role to meshopt. People could expect this to contain an `EXT` meshopt model. They should be able to rely on this
- There is _no_ indication about _which meshopt version_ is used for these models. People could try to load such a model with a loader that supports `EXT`, but that loader will fail, because it's `KHR` now.
- And in hindsight, one could make a case to _not_ assign such a special role to one extension, by using a short form of its name in the directory name. Not sure whether `glTF-SOME_example_extension` would have been better, but ... it is the way it is now.
- More broadly: We'll have to think about a better handling of extensions (and "categorization by extensions") here...

javagl:
That sounds reasonable. 

**Iff** this is done (whoever is responsible for the final 👍 or 👎 on that...), then the directory from https://github.com/KhronosGroup/glTF-Sample-Assets/pull/257 should also be called `glTF-Meshopt-KHR` for consistency.

zeux:
@UX3D-haertl Thanks! This is because my branch was based on an older version that used JPG files but they were later changed to PNG. The resulting conflict is silent as the relevant portions of the file didn't change.

I will update this PR to rectify that and also rename the old files to glTF-Meshopt-EXT.

Issue #261, opened 2026-01-03
Wrong names for `LightVisibility` and `CubeVisbility`


javagl:
The `LightVisibility` and `CubeVisibility` assets have a wrong `name` entry in their `metadata.json`.

- The [`name` of `LightVisibility`](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/a39304cad827573c60d1ae47e4bfbb2ee43d5b13/Models/LightVisibility/metadata.json#L20) is "Light Visiblity test for KHR_node_visibility"
- The [`name` of `CubeVisibility`](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/a39304cad827573c60d1ae47e4bfbb2ee43d5b13/Models/CubeVisibility/metadata.json#L20) is "Cube Visiblity test for KHR_node_visibility"

The `name` must be the ~"actual/general/global name" of that asset, _verbatim_, i.e. it must only be `"LightVisibility"` and `"CubeVisibility"`, respectively.

---

@aaronfranke This slipped though during the review of https://github.com/KhronosGroup/glTF-Sample-Assets/pull/235 because

1. it is not documented(!) and
2. the effect (i.e. the resulting issue) is not obvious

I also didn't have on the radar that deviating from this rule would break downstream pipelines - maybe because there haven't been downstream pipelines until recently. But I just noticed that https://github.khronos.org/glTF-Assets/ chokes on these assets:

<img width="656" height="580" alt="Image" src="https://github.com/user-attachments/assets/515b30dd-33e8-49ab-8d5d-a052b21735fb" />

This moves the task to straighten out and update the submission guidelines (c.f. https://github.com/KhronosGroup/glTF-Sample-Assets/issues/39 ) a bit higher on my TODO list. (I already started doing things like https://github.com/KhronosGroup/glTF-Sample-Assets/pull/254 , but will have to dive deeper into that PHP script first...)

javagl:
Fixed as of https://github.com/KhronosGroup/glTF-Sample-Assets/pull/262 (other cleanups, e.g. adding some better documentation to the submission guidelines, are independent of this _specific_ issue)

PR #262, opened 2026-01-06
Fix cube and light visibility metadata names

aaronfranke:
See https://github.com/KhronosGroup/glTF-Sample-Assets/issues/261


javagl:
Thanks @aaronfranke (I opened the issue mainly to keep track of that, and would have fixed it if/when I find the time for ~"a more general cleanup of some things", but it's good to have this off the plate)

PR #269, opened 2026-03-10
Volume Scattering Test Assets

MiiBond:
Here are a couple of possible conformance assets to test KHR_materials_volume_scatter.

The first one test varying values for anisotropic scattering:
screenshot_raytracer

The second one tests varying scattering colours and attenuation distances while attenuationColor is held constant. As multiscatterColorFactor tends to black, scattering is replaced by attenuation.
screenshot_raytrace

This one tests the scatterStrength parameter for thin-walled scattering.
scatterStrenthThin

And this one tests scatterStrength for volumetric scattering.
scatterStrength


MiiBond:
Here are renders in Babylon.js:

screenshot_rasterizer
screenshot_rasterizer

MiiBond:
I've updated the second asset now to vary `multiscatterColorFactor` from yellow to black on the X-axis and vary `attenuationDistance` on the Y-axis. When scattering is black, there will no longer be any scattering. However, the `attenuationColor` is constant for every material and so the scattering will be replaced by absorption as the `multiscatterColorFactor` tends to black.

I also added text labels for the axes.

emackey:
During the PBR meeting we talked about the need for updated screenshots here.

Models/ScatterAnisotropy/metadata.json
```diff
@@ -0,0 +1,25 @@
...
+            "text": "CC BY 4.0 International",
+            "spdx": "CC-BY-4.0",
+            "icon": "https://licensebuttons.net/l/by/4.0/88x31.png"
+        }
+    ],
+    "tags": [
+        "showcase",
+        "extension"
+    ],
+    "screenshot": "screenshot/screenshot_raytrace.png",
```
emackey: Looks like the filename is missing an `r` on the end, `screenshot_raytracer.png`.

That's why the CI failed.

MiiBond: Good catch. Fixed.

MiiBond:
I've updated the scattering assets to use the new KHR_materials_scatter extension as well as added a couple of new assets to test the new `scatterStrengthFactor` and `scatterStrengthTexture` parameters.

mklefrancois:
Here are the current results from [vk_gltf_renderer](https://github.com/nvpro-samples/vk_gltf_renderer)

ScatterColorAndDensity
ScatterAnisotropy
ScatteringSkull
ScatterStrength
ScatterStrengthThinWalled

emackey:
This branch needs a merge or rebase from main so that it can run the new CI.

Also we talked in the PBR TSG meeting about the need for the text labels on "Scatter Anisotropy" to avoid using +/- 1.

Models/ScatterAnisotropy/metadata.json
```diff
@@ -0,0 +1,25 @@
+{
+    "version": 2,
+    "legal": [
+        {
+            "license": "CC-BY 4.0",
```
emackey, line 5: I think all four of the `metadata.json` files in this PR need to use dashes here, `CC-BY-4.0` instead of the space. That appears to be the cause of the CI failure.

MiiBond: Ah, thanks!

PR #273, opened 2026-04-16
Transmission tinting asset

MiiBond:
This asset is meant to test a renderer's ability to display surface tinting for thin-walled and volumetric materials using either KHR_materials_transmission or KHR_materials_diffuse_transmission.

Ideally, the rendered surface tinting will account for the number of faces that the light actually passes through. i.e. the simple plane should appear less tinted than the cubes.

This asset is also a good test for renderers that do not support the surface-tinting capabilities of the glTF material model directly and must convert these features. e.g. OpenPBR and Adobe Standard Material. In these cases, features like coat tinting might be needed to achieve the same effect.


emackey:
This asset deliberately tests what happens when KHR_materials_diffuse_transmission is applied on top of a volume. This test effectively helped sway the PBR TSG members to believe that the effect should not be applied to a volume, leading to diffuse transmission functionality being merged into a newer KHR_materials_scatter extension that explicitly avoids doing this to volumes of material.

Although this asset is beautiful, it would seem it has now served its purpose, and can be retired as we work to migrate the good parts of the functionality to the new extension.

Any objection to closing this?

PR #275, opened 2026-04-24
Update Scattering Skull asset with newest spec

MiiBond:
This PR just updates the `multiscatterColor` parameter to `multiscatterColorFactor` which matches the most recent spec of `KHR_materials_volume_scatter`


emackey (approved):
Thanks!

PR #279, opened 2026-05-05
Fix inverted-exterior duplicate triangles in Corset model

### LINKED ISSUE #278: Corset Model Contains a Flipped Polygon in Front of the Surface
The Corset model appears to contain an incorrectly placed interior polygon that protrudes through, or lies in front of, the visible exterior surface. During path-traced rendering, the camera ray hits the exterior front-facing surface first, so the misplaced polygon is not visible directly. However, shadow rays still intersect this flipped or internal polygon, causing an incorrect shadow patch on the model.

The following visuals were rendered with [vk_gltf_renderer](https://github.com/nvpro-samples/vk_gltf_renderer).

| Front/Back faces | Rendered | 
| --- | --- |
|Image|Image|

bghgary:
## Context

The Corset model contains a large number of "inverted-exterior duplicate" triangles — triangles whose vertex positions lie on the model's outer surface, but whose face normals point *inward* (toward the model's central axis), coincident with properly outward-facing triangles right next to them. Most rasterizing viewers hide these via backface culling and z-fight resolution, so they have been invisible in tools like the Babylon Sandbox, model-viewer, and the Khronos Sample Viewer's rasterizer mode.

Path tracers don't cull during BVH traversal, and they fire shadow rays directly through the BVH. The phantom interior surfaces win the BVH-traversal coin flip on some pixels (corrupting shading, since the surface gets classified as back-facing) and intercept shadow rays cast from properly-lit points (producing phantom shadows). Issue #278 reported one example of the latter: a hard dark patch on the upper-right chest in `vk_gltf_renderer`'s path tracer.

## Root cause

For a representative example, glTF triangles **7520** and **7521** form a single source-FBX quad (`Models/Corset/sourceModels/Corset.fbx` polygon `#3798`, vertices `[3956, 3860, 3864, 3955]`). Both triangles' positions sit on the proper outer surface, but their face normals point inward and their vertex normals were authored to match (so winding-only checks don't flag them). They duplicate the exterior surface tri-for-tri, just with the winding reversed.

Surveying the whole mesh, **3,490** triangles match this pattern (face normal points radially inward, and a properly-oriented outward-facing triangle exists at the same radius within ≤2 mm). That's roughly 19% of the mesh's 18,324 triangles. The remaining inward-facing triangles (~30) sit at smaller radius than the outer shell and are legitimate inner-liner geometry; they were preserved.

## Fix

Removed the 3,490 inverted-exterior duplicate triangles from the index buffer. The proper outward-facing triangles already cover the model surface, so removal does not leave holes. All three variants were regenerated:

| Variant | Before | After | Notes |
|---|---|---|---|
| `Corset/glTF/Corset.bin` | 662,184 B | 641,244 B | indices bufferView + accessor count updated |
| `Corset/glTF/Corset.gltf` | 18,324 tris | 14,834 tris | accessor count `54972 → 44502`, bufferView byteLength `109944 → 89004`, buffer byteLength `662184 → 641244` |
| `Corset/glTF-Binary/Corset.glb` | 13,491,364 B | 13,471,000 B | regenerated from fixed glTF with `gltf-pipeline` |
| `Corset/glTF-Draco/Corset.bin` | 230,116 B | 215,792 B | regenerated from fixed glTF with `gltf-pipeline -d --separate` |

The `glTF-Binary` and `glTF-Draco` variants were regenerated using [`gltf-pipeline`](https://github.com/CesiumGS/gltf-pipeline) — the same tool the upstream Corset Draco file was originally encoded with (see [`KhronosGroup/glTF-Sample-Models@7f3cd41`](https://github.com/KhronosGroup/glTF-Sample-Models/commit/7f3cd4111808e2282b704d79f9adaf79ce19d9eb)). The Draco bitstream version (2.2) and encoder method (edgebreaker) are byte-identical between the upstream and regenerated files; only the compressed payload differs.

## Verification

- **Khronos glTF Validator (2.0.0-dev.3.10):** all three variants pass (`Errors: 0, Warnings: 0`).
- **vk_gltf_renderer path tracer:** rendered the fixed asset from a camera matching the issue's reported angle. The dark diamond on the upper-right chest is gone. Diff vs the original at the same camera shows changes only in the affected region (and faint AO tightening elsewhere where shadow rays no longer punch through the duplicate shell).
- **Default-camera path tracer render:** no missing geometry, no new artifacts.
- **Rasterizer comparison (Babylon Sandbox):** the rasterized image is unchanged before vs after the fix. This is expected — backface culling was hiding these triangles all along — and confirms no rasterization regression.

## Source FBX

The same inverted-exterior duplicates exist in `glTF-Sample-Models/sourceModels/Corset/Corset.fbx`. Updating the source FBX is outside the scope of this PR; this fix only touches the glTF-format derivatives.

Closes #278

[Created by Copilot on behalf of @bghgary]


javagl:
@bghgary It's not really trivial to verify this (without setting up a path tracer test locally). But I assume that this was tested sufficiently so that it could be merged, right?

emackey:
This branch needs a merge or rebase from main so that it can run the new CI.

emackey:
image

Main on the left, this branch on the right.  With the back faces removed, there can be new rendering artifacts, seen here in the upper-right of this screenshot.  Although the new model has fewer vertices and fewer triangles, this is not the correct solution.

See my comments in https://github.com/KhronosGroup/glTF-Sample-Assets/issues/278#issuecomment-5411711520.  I think this PR should be closed.

bghgary:
Fair enough. We can discuss whether to keep the model or not in 278.

PR #277, opened 2026-05-01
Update REUSE and Licenses

KhronosWebservices:
These changes are made to help ensure we have better coverage on copyright/license, and to clarify what is expected on contributions where AI may be involved:

- Migrate REUSE to current version
- Renamed SubmittingModels.md -> CONTRIBUTING.md to fall in line with other repositories in KhronosGroup
- Added AI Contribution text to CONTRIBUTING.md
- Fixed typos breaking REUSE lint (CC BY-SA-NC 4.0 -> CC-BT-SA-NC-4.0)
- Added CC-BT-SA-NC-4.0 license
- Add missing DGG license, using standard used from other 3rd party model licenses

Issue #49, opened 2022-10-11
How should different screenshot sizes be handled?


javagl:
Inspired by a [comment in a PR](https://github.com/KhronosGroup/glTF-Sample-Models/pull/360#discussion_r992280311):

There is a convention to add a file `/screenshot/screenshot.*` for each model. This screenshot is used as a preview in the table in the main `2.0/README.md` file. 

There are some samples where the screenshot can be pretty small - for example, all the basic `Simple...` models and others. For other screenshots that show nice rendering features, there should be _at least_ one large screenshot that shows the rendered result in high resolution. (Some models already do that, by adding a `screenshot.*` for the preview, and a `screenshot_large.*` to be embedded in the respective README). 

1. Should the `/screenshot/screenshot.*` file be _explicitly_ required to be a "small" one, with low resolution and file size? 
   I think that one could argue that it should be small, so that the preview page is loaded quickly.
2. Should the screenshots in the table be fixed to a certain width? 
   I think that the "jagged" display that results from different-sized screenshots is not so pretty. As an example, showing one table once in [the current form](https://github.com/javagl/glTF-Sample-Models/tree/master/2.0#showcase-1) and once with [fixed image widths](https://github.com/javagl/glTF-Sample-Models/tree/table-column-width/2.0#showcase-1): 

  ![Khronos glTF Screenshot sizes](https://user-images.githubusercontent.com/5597569/195097947-aec02768-7718-43cb-8229-6b9f36ea11ac.gif)

  (This may raise questions about the desired _aspect ratio_ of the screenshots, but I think that fixing the _width_ could make sense...)

javagl:
As of https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280 , the [current state of `CONTRIBUTING.md`](https://github.com/javagl/glTF-Sample-Assets/blob/d2996d653362f1b3b22129bd9c1d24d9f3957f5a/CONTRIBUTING.md#the-screenshot)
 says

> The screenshot should have a width of roughly 150 to 500 pixels, and not be extremely elongated in one direction.

which is about as specific as we can get here. The listing files that show the tables now include the images as
`<img src="..." width="250px">`
i.e. with a fixed width. 

I'll close this one. Feel free to reopen if you disagree.

PR #280, opened 2026-05-21
Update 'utils' from PHP to TypeScript

javagl:
## Previous state

The 'utils' directory contained two PHP scripts, `model.php` and `modelmetadata.php`. The purpose of these scripts was to create and update several files in the repository. For example:

- The `model-index.json` that lists all models and their tags and variants
- The "listing" files like `Model-showcase.md` that shows all models with the 'showcase' tag
- The 'dep5' file that contains the repo-wide license information
- For each model: 
  - The `README.md` file 
  - The `LICENSE.md` file
  - The `metadata.json` file

The `metadata.json` file was also the _input_ for all of this, and there have been a few functionalities that appeared to be related to various forms of "updating" the license information (that have been a bit hard to follow, admittedly).

## New state

The (relevant?) functionality of these scripts is rewritten in TypeScript here. This includes the introduction of structures/types for ... "various things", extensive documentation, and (what I hope to be) clearer processes.

#### There is a `main.ts` that only parses the command line arguments.

Some of the command line arguments had been marked as obsolete, or they have been "smeared" through the code in a form that made it hard to parse (~40 appearances of `dry-run` or `dryRun`...). Once the command lines are parsed, they are passed to the only "public function", namely `ModelMetadata.process`

#### The `ModelMetadata` class summarizes the main, actual tasks. 

It collects the models that should be processed (usually, all models from the `Models` subdirectory). These are collected as `Model` objects (details below). Once the models are collected, several functions are called:

- `createListings`, which creates the list files that summarize models by their tags, e.g. `Models-showcase.md`
- `createModelIndex`, which created the `model-index.json` file
- `createReuseLicense` which creates the `dep5`-file with license information about all models
- Optional: `updateAllModelsFiles`: This updates the `README.md`, `LICENSE.md`, and `metadata.json` of all models

#### The `Model` class summarizes information about "one model"

This class contains, typed, documented information. The main goal was to avoid lines like
`$this->metadata['createReadme'] = (isset($this->metadata['AutoGenerateREADME'])) ? $this->metadata['AutoGenerateREADME'] && $this->metadata['createReadme'] : $this->metadata['createReadme'];`
and to have a more clearly encapsulated, consistent functionality. This _does_ require a certain level of _validation_. Therefore...:

#### The `Models` class is used for creating `Model` objects.

This class takes the data that was read from the `metadata.json` of the models, and performs several levels of validation:

- It checks that all required fields in the metadata JSON are present
- It checks that all required fields in the `legal` array of the metadata JSON are present
- It checks that the metadata is valid
- It checks that all 'legal' objects are valid

Only when all this succeeds, it creates an actual, valid `Model`. (Otherwise, it collects any errors/warnings, so that the caller can deal with that). 

## Changes to licenses

No, this is not what it sounds like. The point is that the previous state defined an [array of licenses](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/util/modelmetadata.php#L118), and there have been quite a few inconsistencies in that. 

For example, the so-called `spdx` identifiers have not been valid SPDX identifiers. Some of the license links have been plainly wrong. I don't know where [this CC0-1.0 link](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/Models/DiffuseTransmissionTeacup/LICENSE.md?plain=1#L7) is coming from, but it redirects to some random blog post. Also, there is no estabished license that is called "Public Domain". And there is no difference between "CC-BY International 4.0" and "CC-BY-4.0". Long story short: There are now exactly [(EDIT:) **four** licenses](https://github.com/javagl/glTF-Sample-Assets/blob/21dfe8958dcbd6f5cfe06e839e5993e60d426ca9/util/data/licenses.json) left. When a `metadata.json` contains a license where the `license` is **not** a valid, known SPDX identifier, then it is assumed to be a "custom" license, meaning that it must have a `licenseUrl` that points to one of the licenses in this repository.

(The current state here contains a few quirks to "canonicalize" and "clean up" the license information in the _existing_ `metadata.json` files...)

# Review

I know, this is a big change, coming out of the blue (or not so, because we talked about some of that earlier). So I assume that not everybody will be willing or able to review this just so. Nevertheless, I'll tag a few people as potential reviewers here:

@DRx3D @emackey @lexaknyazev 

The process for running this state locally should be:

- `cd util`
- `npm install`
- `npm run check`
- `npm run update`

This will check/update all files locally. And the result may look faaar worse than it actually is: It will mark (nearly) all files as "modified", but nearly all these modifications are only the result of canonicalizing the license names to their proper SPDX names. For the other files, I tried to keep the structure and layout the same as before.


util/data/listings.json
```diff
@@ -0,0 +1,42 @@
+[
+  {
+    "file": "Models.md",
+    "tags": [],
+    "summary": "All models listed alphabetically."
+  },
+  {
+    "file": "Models-core.md",
+    "tags": ["core"],
+    "summary": "Models that only use the core glTF V2.0 features and capabilities."
```
lexaknyazev: nit: we usually call it "glTF 2.0", without "V".

util/src/main.ts
```diff
@@ -0,0 +1,150 @@
...
+  },
+  {
+    long: "check",
+    short: "c",
+    text: "Checks consistency of the asset directory files.",
+  },
+  {
+    long: "update",
+    short: "u",
+    text: 'Update model folders. It has no effect "check" fails. Will set "check".',
```
lexaknyazev: Seems like some language is either missing or incorrect in this description.

util/src/main.ts
```diff
@@ -0,0 +1,150 @@
...
+ * @param argv - The arguments
+ * @returns The parsed command line
+ */
+function parseCommandLine(argv: string[]): ParsedCommandLine {
+  const result: ParsedCommandLine = {
+    options: {},
+    modelNames: [],
+  };
+
+  //console.log("argv ", argv);
```
lexaknyazev: If this is useful for debugging, it should probably be conditioned on some env or compile flag.

util/src/main.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  // Note: 'argv[0]' and 'argv[1]' are the script names!
+  for (let i = 2; i < argv.length; i++) {
+    const arg = argv[i];
+    let wasOption = false;
+    for (const option of commandLineOptions) {
+      if (arg === `--${option.long}` || arg === `-${option.short}`) {
```
lexaknyazev: It might be better to disallow the same option being used both in short and long forms at the same time.

util/src/main.ts
```diff
@@ -0,0 +1,150 @@
...
+        result.options[option.long] = true;
+        wasOption = true;
+      }
+    }
+    if (!wasOption && arg !== undefined) {
+      result.modelNames.push(arg);
```
lexaknyazev, line 89: This condition implies that a model name can start with one or two dashes as long as it does not match any of the options exactly. It also implies that it's possible to interleave options and model names.

Not sure if those cases need to be forbidden.

By the way, when could `arg` be undefined here?

util/src/Model.ts
```diff
@@ -0,0 +1,593 @@
...
+
+    const directoryEntries = fs.readdirSync(modelDirectory, {
+      withFileTypes: true,
+    });
+    for (const directoryEntry of directoryEntries) {
+      const name = directoryEntry.name;
+      if (directoryEntry.isDirectory() && name.startsWith("glTF")) {
+        const directory = `${modelDirectory}/${name}`;
+        const files = fs.readdirSync(directory);
+        for (const file of files) {
```
lexaknyazev: It might be useful to issue an error if a file uses a wrong casing for a known extension, e.g., `.glTF`.

util/src/Model.ts
```diff
@@ -0,0 +1,593 @@
...
+  /**
+   * Returns the URL of the screenshot.
+   *
+   * This is getScreenshotPath, URL-encoded
+   *
+   * @returns The screenshot URL
+   */
+  getScreenshotUrl() {
+    const p = this.getScreenshotPath();
+    const result = p.replace(/ /g, "%20");
```
lexaknyazev: Why not `encodeURI(p)`?

util/src/Model.ts
```diff
@@ -0,0 +1,593 @@
...
+  /**
+   * Returns the screenshot name that was read from the metadata JSON
+   *
+   * This is getScreenshotName, URL-encoded
+   *
+   * @returns The screenshot name
+   */
+  getScreenshotNameUrl() {
+    const p = this.getScreenshotName();
+    const result = p.replace(/ /g, "%20");
```
lexaknyazev: Ditto.

util/src/Model.ts
```diff
@@ -0,0 +1,593 @@
...
+  /**
+   * Returns the URL of the ".glb"" file for the "glTF-Binary" variant.
+   *
+   * This is getGlbPath, URL-encoded
+   *
+   * @returns The GLB URL
+   */
+  getGlbUrl(): string {
+    const p = this.getGlbPath();
+    const result = p.replace(/ /g, "%20");
```
lexaknyazev: Ditto.

util/src/Model.ts
```diff
@@ -0,0 +1,593 @@
...
+  /**
+   * Returns the URL of the ".gltf" file for the "glTF" variant.
+   *
+   * This is getGltfPath, URL-encoded
+   *
+   * @returns The glTF URL
+   */
+  getGltfUrl(): string {
+    const p = this.getGltfPath();
+    const result = p.replace(/ /g, "%20");
```
lexaknyazev: Ditto.

util/src/Model.ts
```diff
@@ -0,0 +1,593 @@
...
+      md.push("### Required");
+      md.push("");
+      md.push(this.extensionsRequired.map((e) => "* " + e).join("\n"));
+      md.push("");
+      md.push("### Used");
+      md.push("");
+      md.push(this.extensionsUsed.map((e) => "* " + e).join("\n"));
+      md.push("");
+    } else {
+      if (this.extensionsRequired.length > 0) {
```
lexaknyazev: Since all required extensions must be used, this branch is unreachable.

Besides, why should the headings structure be different depending on required extensions presence?

util/src/Model.ts
```diff
@@ -0,0 +1,593 @@
...
+   *
+   * @returns The lines
+   */
+  createCreditsMarkdownLines(): string[] {
+    const credits: string[] = [];
+
+    const metadata = this.metadata;
+    for (const element of metadata.legal) {
+      const url = element.licenseUrl ?? "";
+      const urlReplaced = url.replace(/ /g, "%20");
```
lexaknyazev: Why not `encodeURI`?

util/src/Model.ts
```diff
@@ -0,0 +1,593 @@
...
+    );
+
+    // The "text" (short name) of the licenses, linking to the
+    // license URL
+    for (const legal of this.metadata.legal) {
+      let licenseUrl = legal.licenseUrl;
+      if (licenseUrl === undefined) {
+        licenseUrl = "";
+      }
+      licenseUrl.replace(/ /g, "%20");
```
lexaknyazev: Ditto.

util/src/ModelMetadata.ts
```diff
@@ -0,0 +1,551 @@
...
+   * The URL for the glTF-Sample-Viewer release
+   */
+  static readonly UrlSampleViewer =
+    "https://github.khronos.org/glTF-Sample-Viewer-Release/";
+
+  /**
+   * The URL of the "raw" glTF-Sample-Assets repo
+   */
+  static readonly UrlModelRepoRaw =
+    "https://raw.GithubUserContent.com/KhronosGroup/glTF-Sample-Assets/main";
```
lexaknyazev: Domain names are case-insensitive.

util/src/ModelMetadata.ts
```diff
@@ -0,0 +1,551 @@
...
+
+    // Create all models
+    const modelIssues: Record<string, Issues> = {};
+    const models = Models.createModels(baseDirectory, modelNames, modelIssues);
+
+    // If there are any errors (or if 'check' was requested),
+    // print all issues.
+    let anyHasErrors = false;
+    for (const i of Object.values(modelIssues)) {
+      anyHasErrors = anyHasErrors || i.errors.length > 0;
```
lexaknyazev, line 73: nit: might exit the loop on the first non-zero `i.errors.length` value or, likely better, combine this loop with the next one to iterate only once.

util/src/ModelMetadata.ts
```diff
@@ -0,0 +1,551 @@
...
+      row.push(`${credits.join("<br>")}`);
+      row.push(` |`);
+
+      md.push(row.join(""));
+    }
+
+    // Footer
+    md.push("---");
+    md.push("");
+    md.push(`### Copyright\n\n&copy; 2023, The Khronos Group.`);
```
lexaknyazev: 2026 (or better use the current year)

util/src/ModelMetadata.ts
```diff
@@ -0,0 +1,551 @@
...
+      modelIndex.push(modelJson);
+    }
+    if (ModelMetadata.verbose) {
+      console.log(`Creating model index for ${models.length} models DONE`);
+    }
+    return modelIndex;
+  }
+
+  /**
+   * Create the "./.reuse/dep5" file that aggregates the license information
```
lexaknyazev: DEP5 is deprecated as of the current REUSE spec. The replacement file is `REUSE.toml`.

util/src/ModelMetadata.ts
```diff
@@ -0,0 +1,551 @@
...
+      "Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/"
+    );
+    dep.push(
+      "Source: glTF V2.0 models from various sources collected into a Repo"
+    );
+    dep.push("Upstream-Name: glTF V2.0 Model Repo");
+    dep.push(
+      "Upstream-Contact: https://GitHub.com/KhronosGroup/glTF-Sample-Models/"
+    );
+    dep.push("Copyright: 2017-2023 Khronos Group");
```
lexaknyazev: 2026 or better the current year, also below.

util/src/ModelMetadata.ts
```diff
@@ -0,0 +1,551 @@
...
+      console.log(`Creating dep5 for ${models.length} models...`);
+    }
+
+    const dep = [];
+
+    dep.push(
+      "Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/"
+    );
+    dep.push(
+      "Source: glTF V2.0 models from various sources collected into a Repo"
```
lexaknyazev: Should be "glTF 2.0", also below.

util/src/Models.ts
```diff
@@ -0,0 +1,357 @@
...
+  /**
+   * Do some updates of legacy stuff...
+   *
+   * @param metadataJson - The metadata JSON
+   */
+  private static updateLegacyLicenseStructure(metadataJson: any) {
+    const knownLicenses = Licenses.LICENSE;
+    const legals = metadataJson.legal ?? [];
+
+    const legacyLicenseNames: Record<string, string> = {
```
lexaknyazev: We should not allow legacy names for new submissions so this block is going to be removed, right?

util/src/main.ts
javagl: I'm currently going through the comments and resolving the 'trivial' ones once I fixed them locally, and will push an update for this soon.

I tried to keep some compatibility/consistency with the PHP version (which is also the reason for the odd wordings in the 'descriptions' mentioned above). For this _particular_ request: I'd rather omit the possibility to even _use_ the 'short' options. One can go very far with command line parsing, but in this case, the scope is very limitied, so I'd rather keep it simple instead of trying to add consistency check like this. 

Opinions on omitting the 'short' options?

util/src/main.ts
javagl: Similar to above: I tried to mimic the command line parsing of the PHP version.

The `arc` cannot be `undefined` here, but TypeScript does not know this. Its type is `string | undefined`, which cannot be pushed into the names.

util/src/main.ts
lexaknyazev: Omitting short options sounds good.

util/src/main.ts
lexaknyazev: That could likely be fixed by declaring `arg` like this on L93:
```ts
const arg = argv[i]!;
```

javagl: There are contexts/settings in which this is discouraged, via things like https://typescript-eslint.io/rules/no-non-null-assertion/ , because it _can_ break the little type-safety that TypeScript offers _in theory. No strong opinion, though, so changed it...

util/src/Model.ts
javagl: In principle, it should be allowed to have uppercase letters there. But I agree that it is unusual enough to issue a warning, at least. Maybe more importantly: There should be a check of the case that _no_ file (neither glTF nor GLB) is found for a variant. So I extended this check: The `initialize->findVariants` call is now done from the outside, receives `Issues` to track, and adds

- a warning for non-standard upper/lowercase
- an error when no matching glTF/GLB file is found

If it really should be an error, feel free to reopen

util/src/ModelMetadata.ts
javagl: This is from the PHP version. I had seeen the uppercase letters. I wondered. And I thought: "Domain names are case-insensitive", and left it at it was. But now, I changed it to lowercase.

util/src/Model.ts
javagl: For this and the ones mentioned below: There has been a bit of back and forth in the PHP version about this. There have been ["file names" and "paths" and "Uris"...](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/util/modelmetadata.php#L732). In some places, the `%20` was inserted, and in some other places, it was even replaced by ` ` again as a "backwards translation". 

I _tried_ to somehow straighten this out, with the specific, named, documented `...Path` and `Url` flavors of these functions here, but it may warrant another cleanup pass.

I now changed this "replace"/`%20` _everywhere_ to `encodeUri`. This does change the links in the READMEs, e.g.
`Unicode❤♻Test` to
`Unicode%E2%9D%A4%E2%99%BBTest`
But ... I think that this is _at least_ OK, and maybe even _intended_, for the cases of non-URI-compatible characters.

util/src/Model.ts
javagl: That layout was taken from https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/util/modelmetadata.php#L435 . I roughly tried to "keep the layout the way it is". However, I now changed it so that both sections use
`## Extensions ....`
if they are present.

util/src/ModelMetadata.ts
javagl: Yeah, this was _somewhat_ restructured from the [original `check`](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/util/model.php#L90) implementation, and iterating _twice_ indeed felt a bit awkward. But the intention here was to print all errors when there is _any_ error (or `check` was given), to avoid the case where you have to run `check` 10 times, only to receive _one_ (new) error report in each run. 

In pactice, this script will be run mainly in CI, for _one_ added model, but I still think that reporting all errors should make sense. 

I'll leave it to you to close or disagree with this one.

util/src/ModelMetadata.ts
javagl: I was hesitating about that initially. I wanted to avoid the case that a new year changes "all files". But given that this is only written into the listing- and dep5-files that should not be an issue.

util/src/ModelMetadata.ts
javagl: Now I'm looking at a 16 year old blog post about how to handle Unicode in Python. A bit more broadly: These "reuse" people really, _really_, **really** missed the mark here, right?

util/src/Models.ts
javagl: YES! This should be a one-time update, essentially. 

(I considered to make this an _explicit_ PR, so that this handling can be omitted as soon as this PR is merged, but ... let's see how that "reuse" mess is sorted out...)

util/src/ModelMetadata.ts
lexaknyazev: As currently written, the code prints all errors and all warnings from all models if a single model has an error. Is that the intended behavior?

javagl:
An inlined comment pointed out that the `./reuse/dep5` file is deprecated, and this should be a `REUSE.toml` in the repository root. After a bit of back and forth, this is updated here accordingly: The process now writes out a new `REUSE.toml` file during the update. I also cleaned up the `LICENSES` directory accordingly, and now, running
`reuse lint` 
locally after the update prints:

> # SUMMARY
>
> * Bad licenses: 0
> * Deprecated licenses: 0
> * Licenses without file extension: 0
> * Missing licenses: 0
> * Unused licenses: 0
> * Used licenses: CC-BY-4.0, CC-BY-NC-4.0, CC-BY-NC-SA-4.0, CC0-1.0, LicenseRef-3DRT-Testing, LicenseRef-Adobe-Stock, LicenseRef-CRYENGINE-Agreement, LicenseRef-LegalMark-Cesium, LicenseRef-LegalMark-DGG, LicenseRef-LegalMark-Khronos, LicenseRef-LegalMark-UX3D, LicenseRef-Poser-EULA, LicenseRef-Stanford-Graphics, SCEA
> * Read errors: 0
> * Invalid SPDX License Expressions: 0
> * Files with copyright information: 2317 / 2317
> * Files with license information: 2317 / 2317
> 
> Congratulations! Your project is compliant with version 3.3 of the REUSE Specification :-)

So that seems to be fine. 

---

The careful observer might have noticed that this PR now removes **19.160** lines. I have no clue what [`reuse.spdx`](https://github.com/javagl/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/.reuse/reuse.spdx) was, but am 100% sure that it should either not have been there at all, or at least, not in the given form. You know, with these [random parts of `.gltf` files](https://github.com/javagl/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/.reuse/reuse.spdx#L13309) and such.

---

The validation via `reuse lint` could (and probably should) be part of the CI one day, but I think that this could be addressed in a dedicated PR.

---

Another inlined comment pointed out the [`updateLegacyLicenseStructure` function](https://github.com/javagl/glTF-Sample-Assets/blob/21dfe8958dcbd6f5cfe06e839e5993e60d426ca9/util/src/Models.ts#L278) that is part of the current state. This step essentially just updates all `metadata.json` files to use proper, canonical SPDX identifiers.

This should **not** be part of the state that is eventually merged. I think that one way of dealing with this would be:

- Run the `update` locally
- Commit the updated state (basically updating _all_ files in the repo, including the `metadata.json` files)
- Remove that `updateLegacyLicenseStructure` function call
- Commit the state without that function

Any thoughts on that...?

util/src/ModelMetadata.ts
javagl: Addressed in the comment below at https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280#issuecomment-4522073671

util/src/Models.ts
javagl: TBD in at the bottom of the comment below at https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280#issuecomment-4522073671

util/src/ModelMetadata.ts
javagl: That was the intention, yes. The PHP state had some `no-warn` flag for

> Do not display warning messages unless there are errors

I don't have strong preferences on some of that, except for that it should print **all** errors when there is **one** error (whether or not any 'warnings' are omitted then is a detail - the only 'warnings' currently are about the "file extensions being not-all-lowercase" anyhow...)

javagl:
Small addendum to the last point from the previous comment: Such a "one-time-cleanup pass" could also fix https://github.com/KhronosGroup/glTF-Sample-Assets/issues/281 . The current state contains some ["debug" function](https://github.com/javagl/glTF-Sample-Assets/blob/8a83839c62b673d447e0a011fc84fc649a03266c/util/src/ModelMetadata.ts#L104) to print these cases. Of course, that will be removed before marking this PR as ready, but only when the intended workflow for that one-time-cleanup is sorted out.

lexaknyazev:
My personal preference would be to split changes into more targeted PRs, e.g., scripts vs data as appropriate, stacked on top of each other or opened sequentially. At the very least, _obvious and trivial_ one-time cleanups (removal of `reuse.spdx` in particular) could be merged separately in advance to not inflate PRs that contain changes for reviews.

That said, other maintainers may have different preferences and, ultimately, the goal is to eventually land all changes, so an all-in-one PR that brings the repo to a new state is great too. Up to you how to organize PR's commit history (it can be rewritten if desired).

javagl:
It's not clear know how that file ended up here, but the step of removing it is just part of the update for the `dep5`-recation. It could _probably_ have been done at any time, as a 1-commit-PR. There could also be a dedicated PR for https://github.com/KhronosGroup/glTF-Sample-Assets/issues/281. And a dedicated PR for that unused `test` tag. And one for the newly added `pbrtest`. We could set up a project board, and have planning meetings, and set up a task force, and define a roadmap, but I'll leave that to the managers. I'm just trying to (spend my spare time to) get some overdue cleanup work done here.

By the way, I just saw https://github.com/KhronosGroup/glTF-Sample-Assets/pull/277. Did anybody care about that? No. Should it have been merged by now? Yes. However, I'll try to align it with the changes here.

lexaknyazev:
> By the way, I just saw #277. Did anybody care about that? No. Should it have been merged by now? Yes. However, I'll try to align it with the changes here.

I suppose there are only two ways of handling this: either merging #277 and resyncing this PR afterwards or closing #277 and incorporating its changes here. As this PR's name suggests that it's about rewriting scripts, merging #277 first would be a bit more semantically accurate.

Anyway, please feel free to structure this work as you see fit. Getting it merged is more important than planning details.

javagl:
> merging https://github.com/KhronosGroup/glTF-Sample-Assets/pull/277 first would be a bit more semantically accurate.

Merging that other PR does not make sense: As soon as it is merged and the PHP script runs, it will build a new `dep5` file, leading to a completely inconsistent state. 

But that other PR is pretty much a _strict subset_ of what is done here: Removing `dep5`, adding `.toml`, adding/fixing some licenses. So in the spirit of not trying to cross a chasm with two jumps, I'd add the other (minor) changes from that PR here (renaming the MD to `CONTRIBUTING.md` and such), and then, **this** PR should supersede the other one.

javagl:
Addressing the points from https://github.com/KhronosGroup/glTF-Sample-Assets/pull/277 

- _Migrate REUSE to current version_
  This was done here as well. In contrast to the other PR, _this_ PR does not only remove the `dep5` file and add the `REUSE.toml` file, but **also** updates the `util`-functionality to actually _generate_ updated `REUSE.toml` files in CI
- _Renamed SubmittingModels.md -> CONTRIBUTING.md to fall in line with other repositories in KhronosGroup_
  This was done here as well. And while I'm at it, and because I have nothing else to do, I essentially re-wrote the `CONTRIBUTING.md`, trying to give it a structure that makes it easier to follow the contribution process: 
  - The document now lists the required files immediately: `metadata.json`, `README.body.md`, screenshot. 
  - It explains **all** properties in the `metadata.json` and its `legal` array, in a way that is hopefully clearer and easier to fill out. 
  - I also added some information about the 'variants' (even though I noticed that the variant names have turned a bit arbitrary...). 
  - I added an appendix, listing the valid 'tags' and licenses (this wasn't really made clear before that)
  - Direct preview for reviewers: [`CONTRIBUTING.md`](https://github.com/javagl/glTF-Sample-Assets/blob/8f1d5a8c8d09081b3b829b34e88589b4f5d94c1b/CONTRIBUTING.md) 
- _Added AI Contribution text to CONTRIBUTING.md_
  Yeah, added an appendix for that as well 🤷 

The remaining points from that PR...

- _Fixed typos breaking REUSE lint (CC BY-SA-NC 4.0 -> CC-BT-SA-NC-4.0)_
- _Added CC-BT-SA-NC-4.0 license_
 - _Add missing DGG license, using standard used from other 3rd party model licenses_

had been done here as well, as part of the overall cleanup of the license namings.

So I think that _this_ PR will make the PR https://github.com/KhronosGroup/glTF-Sample-Assets/pull/277 obsolete now.

---

### Previews:

- A preview for the result of updating all `metadata.json` files to their canonical form is in https://github.com/javagl/glTF-Sample-Assets/pull/1 
- A preview for the intended final state after merging this PR is in https://github.com/javagl/glTF-Sample-Assets/pull/2

javagl:
I'll mark this as "Ready For Review" now (even though I expect some minor cleanups and comments to come in during the review). 

EDIT: This is obsolete now:
<strike>
The only "larger" task that I'd consider to be a remaining TODO would be to add the
`reuse lint`
step to the CI, to ensure that newly submitted models have proper (custom or SPDX) licensing. Some validation _could_ be done by the `utils` here (e.g. checking for the presence of a `.txt` file in the `LICENSES` folder), but maybe that can be left to the 'reuse' linting. 
</strike>

The CI integration of the REUSE check is drafted in https://github.com/KhronosGroup/glTF-Sample-Assets/pull/282 

**Note** Before this is actually merged, the [`updateLegacyLicenseStructure`](https://github.com/javagl/glTF-Sample-Assets/blob/8f1d5a8c8d09081b3b829b34e88589b4f5d94c1b/util/src/Models.ts#L282) function should be removed. (It wouldn't do any harm, but _should_ no longer be necessary in the future)

javagl:
A very specific technical aside:

The `CONTRIBUTING.md` lists the properties that have to go into the `metadata.legal` entries. For SPDX licenses, this does **not** require the properties
```
"licenseUrl": "https://creativecommons.org/licenses/by/4.0/legalcode",
"spdx": "CC-BY-4.0",
"icon": "https://licensebuttons.net/l/by/4.0/88x31.png"
```
These are inserted automatically. I don't know what the intention was behind that `legal` structure, and why it includes that unused `icon`, for example. I think that there could/should be a structure of that `legal` that does _not_ include that icon, but that would be a 'breaking change'. If people think that this should/could be omitted, I'm all in favor of making that `metadata.json` as lean and clean as possible.

EDIT: When I said that this would be a 'breaking change', then this referred to things like https://github.khronos.org/glTF-Assets/model/AnimatedTriangle , which already _does_ display this 'icon' at the bottom, and one has to _assume_ that the URL for that is pulled from the metadata. So it can not simply be removed...

javagl:
Recreational saturday evening coding: The new structure would allow some additional functionalities more easily. For example, it took only a few minutes to extend the functionality of creating 'listings' to other categories - specifically, the _extensions_ that the models are using. A preview: [A listing of all models using `EXT_materials_clearcoat`](https://github.com/javagl/glTF-Sample-Assets/blob/94512a5a61e77c8219fae8cd89f9864e2f9ac393/Models/Models-KHR_materials_clearcoat.md) (with links to all other extension listings).

But of course, at some point, it does not make sense to maintain these as MD files here. The https://github.khronos.org/glTF-Assets/ website (or https://asset-explorer.needle.tools/ ) may be more suitable for all this.

util/src/ModelMetadata.ts
lexaknyazev: Printing everything including warnings if there's a single error is why two loops are unavoidable. If that behavior is intended, there's nothing to change here except adding a comment explaining that.

If we don't need to print warnings in a case of a single error, this part could be expressed simpler and clearer with a single loop.

util/src/Model.ts
```diff
@@ -0,0 +1,616 @@
...
+            if (!file.endsWith(".gltf")) {
+              warnings.push(
+                `File does not have an all-lowercase extension: ${file}`
+              );
+            }
+            variants[name] = file;
+            foundFile = true;
+            break;
+          }
+        }
```
lexaknyazev: Please consider deduplicating this section as  `.endsWith(".glb")` and `.endsWith(".gltf")` can't be true simultaneously.

README.md
```diff
@@ -91,4 +92,4 @@ To make this repository cleaner, the previous _glTF-Sample-Models_ repository wa
 For the assets that remained, there were no changes to the Asset name. READMEs were provided to assist in navigation to any directories that were removed or substantially changed.
 
 ---
-&copy; 2023, The Khronos Group. Licensed as CC-BY 4.0 International
+&copy; 2023, The Khronos Group. Licensed as CC-BY-4.0 International
```
lexaknyazev: Changes to the file seem to warrant updating the year.

CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
+<p align="center">
+<img src="https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/glTF_RGB_June16.svg" height="100">
+</p>
+
+# Managing Models in this Repository
+
+We are actively accepting new or revised models for the Sample Asset repo. This keeps the repo active and relevant. All changes, whether they are new models or updates of existing ones, are handled as Pull Requests (PR) and must meet the same quality requirements.
```
lexaknyazev: Let's align with the repo name: "Sample Asset**s** repo"

CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Model Quality Requirements
+
+A summary of the quality requirements for models is given here. Details for submitting new models will be given in the next section.
+
+- Each model and is contained in a subdirectory of the `./Models` directory. This subdirectory contains all model variants and associated files.
```
lexaknyazev: Looks like a typo: "Each model and is"

util/src/ModelMetadata.ts
javagl: Of course, that loop it's not _strictly_ unavoidable, but a bit more broadly: The only 'warning' right now is
`File does not have an all-lowercase extension: ${file}`
You originally suggested to treat this as an error. That's OK for me (and would make the discussion about 'warnings' obsolete, including that `warnings: string[]`). 

Do we want 'warnings' after all? For which cases? Should they always be printed, or always be ignored? If someone has strong opinions on all of that, I could quickly change this. (Otherwise, we might discuss this specific aspect in a dedicated issue)

CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
...
+- Each model must pass the [glTF-Validator](https://github.khronos.org/glTF-Validator/).
+- Each model must have an associated `README.md` markdown file that describes the model and the features of the model that make it appropriate of this repo
```
lexaknyazev: typo: "appropriate for"

CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
...
+- Each model must have an associated `metadata.json` file that includes legal information (ownership, copyright, and license) 
+- Each model must have a properly formatted screenshot
```
lexaknyazev: Maybe explain "properly formatted" here or elsewhere; the dedicated section below does not contain any specifics either.

CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
...
+
+The meaning of the fields:
+
+- `version`: The version for the metadata JSON format. Must be `2`
+- `name`: A name for the model that will be used to refer to be model in titles and link descriptions. This will usually resemble the name of the subdirectory that contains the model. For example, the `name` of a model that is contained in `./Models/AnimatedTriangle` may be `"Animated Triangle"`.
+- `path`: The path to the model, referring to the root of the repository. For example, `"./Models/AnimatedTriangle"`.
+- `summary`: A short summary of the model, to be displayed in tables and overviews. It should usually be a single sentence of short paragraph, and _not_ a full description of the model. For example, a summary might be `"A simple triangle with a rotation animation"`.
+- `screenshot`: The path to a screenshot that should be displayed for the model, suitable for being displayed in an overview table. 
+- `tags`: An array of tags that are used for classifying the model. See the [Tags](#tags) section for the set of tags that are currently supported.
+- `createReadme`: Whether a `README.md` file should be generated automatically for the model. This should usually be `true`.
```
lexaknyazev: When would this be false?

util/src/Model.ts
javagl: Done. That `else` was actually there in the [original code](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/util/modelmetadata.php#L787). 

Also a bit more generally: 

I'd like that "variants search" to be "better" in some ways. But we currently don't have a strict enough definition of what 'variants' are. I tried to summarize a bit of that in the [updated `CONTRIBUTING.md`](https://github.com/javagl/glTF-Sample-Assets/blob/b8cb1ce57e301b3197124b6d273ab7c5d36b2cc6/CONTRIBUTING.md#the-actual-model). But there are oddly specific ones like `glTF-Binary-KTX-ETC1S-Draco` that are not mentioned there explicitly. 

Nearly a detail: The current check does not complain about a `glTF-Binary` subdirectory containing a `.gltf` file and such. 

We should probably consider to _define_ an accepted set of 'variants' somewhere, maybe even including the expected extension _per variant_ (e.g. `glTF-Binary`: `.glb`). That would likely mean to rename some of the existing ones, and be more strict to disallow things like `glTF-ReallyObscureName`. 

Maybe also worth escalating into an issue...?

CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
...
+  "licenseUrl": "",
+  "year": "",
+  "artist": "",
+  "what": "",
+  "owner": ""
+}
+```
+
+- `license`: An identifier for the license, e.g. `"LicenseRef-LegalMark-Khronos"`
+- `licenseUrl`: A link to the license text in the `LICENSES/` directory of the repository. The license text must be a file with the naming pattern _`<license>`_`.txt`. For example, such a license URL could be `"../../LICENSES/LicenseRef-LegalMark-Khronos.txt"`.
```
lexaknyazev: Isn't this field redundant? If the directory is fixed and the file name must match the license identifier, it would be better to omit it and enforce correct names implicitly (by checking that the file exists).

README.md
javagl: The main `README.md` didn't change soooo much, and usually, the copyright year refers to the _first_ publication (but I don't have strong opinions about that, so updated it to 2026 now)

CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
...
+- `text`: A short description or name of the license and what it refers to
+- The remaining fields have the same meaning as for standard licenses (defined in the previous section)
+
+#### The Screenshot
+
+Each model must have at least one associated screenshot. This should be a small image file that is suitable for being displayed in an overview table. The screenshot must be a PNG-, JPG, or (possibly animated) GIF file. It must be in the `screenshot` subdirectory of the model. The path to this screenshot must be given in the `screenshot` property of the `metadata.json`. Additional (larger) screenshots may be contained in the `screenshot` directory, and used in the `README.body.md` file (explained below).
```
lexaknyazev: typo: "PNG-"; might also allow WebP screenshots (especially for animations).

CONTRIBUTING.md
javagl: (I'll close the fixed typo ones and push the update soon. Some of the text (e.g. 'appropritate of') was taken from the original state, but a cleanup pass should clean up _this_ as well)

The "properly formatted" _probably_ referred to the _size_ of the screenshot. The [current state](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/SubmittingModels.md#requirement-details) mentions that the screenshot should be at most 150 pixels wide, but I think that's a bit too small. I considered to add some details about the screenshots having ~"an aspect ratio of roughly 1.0". But there are some broader questions around all of that in https://github.com/KhronosGroup/glTF-Sample-Assets/issues/49 . If there is any preference (e.g. allowing widths of 150...500, and using a fixed width of `250px` in the table or so), then I could add this here as well.

util/src/Model.ts
lexaknyazev: > That else was actually there in the original code.

I meant deduplicating code blocks themselves as they are identical except for the conditions. That surely could be done later.

> We should probably consider to define an accepted set of 'variants' somewhere

For starters, we should require standard names for standard variants:
- `glTF` and `glTF-Embedded`: `.gltf`
- `glTF-Binary`: `.glb`

All others could be ignored for now.

CONTRIBUTING.md
javagl: This is taken from the PHP state. The only reason for this being `false` is probably when someone wants to create the _whole_ `README.md` manually, and **not** let it be auto-generated from the screenshot and the `README.body.md`. 

I'd actually lean towards removing this `createReadme` flag completely, and _require_ the `README.body.md`, to ensure some consistency between the READMEs, and avoid that ["None provided"](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/Models/Suzanne/README.md#description) that is/was inserted by default (even though _this_ model _does_ have a `body` file - there are some inconsistencies here...)

Anyone opposed to ...

- removing the `createReadme` flag
- requiring the `README.body.md`

?

lexaknyazev: > removing the `createReadme` flag

+1; the main `README.md` should always be auto-generated.

> requiring the `README.body.md`

+1; if there's absolutely nothing to say about the asset, then it should not be added.

CONTRIBUTING.md
javagl: Again, there are some legacy aspects [inherited from the PHP](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/2bac6f8c57bf471df0d2a1e8a8ec023c7801dddf/util/modelmetadata.php#L616) here. 

The current state contains
`"spdx": "LicenseRef-LegalMark-Khronos"`
in metadata files, which I'd consider to be undesirable - it's simply not an SPDX identifier. My intention to clean this up somehow became obsolete: Apparently, the [`REUSE.toml` contains a `SPDX-License-Identifier`](https://github.com/javagl/glTF-Sample-Assets/blob/c9f8adc33c6390b73e20749acd454a4740e30c97/REUSE.toml#L79) that **has** to be a string that is **not** an SPDX identifier - ouch 🤕 

The current intention is that:

- the `license` can _either_ be a known SPDX identifier, or that `LicenseRef...` string
- if it is a known SPDX identifier, then the associated fields (`licenseUrl`, `spdx`, `icon`) are filled out automatically
- if it is a custom license name, then there must be the `LicenseUrl`. Apparently, the new REUSE policy is that there must be that 1-to-1 correspondence between that `license` name, and the `licenseUrl` pointing to the `LICENSES/....txt` file. So yes, this could _probably_ be omitted.

Checking that this file exists is a [TODO in the current state](https://github.com/javagl/glTF-Sample-Assets/blob/18b01be4fc242383e5f697d7d79478820bad23e7/util/src/Models.ts#L268). This existence check is _also_ done by the `reuse lint` step. But it could be better to catch this earlier in the process (i.e. at the place marked with that `TODO`). 

I'll try to remove the `licenseUrl` and check the validity earlier. (There currently are some places with a wrong `../`-nesting, so I have to review that anyhow)

util/src/ModelMetadata.ts
lexaknyazev: Let's treat that as an error. On case-sensitive file systems one could put two files like `asset.gltf` and `asset.glTF` in the same directory (may be subject to git config) and that would cause even more confusion in the script.

CONTRIBUTING.md
javagl: This was addressed in https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280/commits/b7a00dca3dd55c502ee9439b5b2dab12f06da0e9

CONTRIBUTING.md
javagl: This was addressed in https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280/commits/8743f378d39fe14c3d0ea1be85519ecf8a2e5212

util/src/Model.ts
javagl: This was added via https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280/commits/eb8672e2624f3f672e0d1c0a4ba0bce372546237 (only a slight improvement, but maybe OK for now)

util/src/ModelMetadata.ts
javagl: I had seen that comment later, after addressing https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280#discussion_r3299327379 (and now fixed this with https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280/commits/49de53f5badd3a94298549a6180cdae5f7e9b256 )

So strictly speaking, the `warnings` are now unncessary. If we decide that there will never be 'Warnings', they could be removed completely.

javagl:
@DRx3D When you have time... : I think that we could try to get this merged sooner or later

DRx3D:
I admire those who can review any large PR unassisted. I used Claude to help me find a few things. Major items will get their own issue, and minor ones will be bundled together. I am completely in favor of this work so my comments are not meant to derail it, but help if get completed, better. Any of them can be pushed off to a future release if it makes more sense to do that.

@javagl explicitly describes these as "must remove before merge"
1. Models.updateLegacyLicenseStructure() (util/src/Models.ts, called unconditionally from createModel) — a one-time metadata.json migration shim. Author said explicitly: "This should not be part of the state that is eventually merged" and later "Before this is actually merged, the updateLegacyLicenseStructure function should be removed." It's still there, still called on every run.

1. Dead code: ModelMetadata.DEBUG_preprocessModels() (util/src/ModelMetadata.ts) — //@ts-ignore-marked debug dump of all extensions/tags/variants. Author: "that will be removed before marking this PR as ready." Still in the file; only call site is commented out, so it's dead code.

DRx3D:
Screenshots have never been formally defined. It was thought that a minimum width of 150px would work. But there are tall vertical images (stained glass lamp) and wide horizontal images (GlamVelvetSofa). The repo UI is (or should be) handled by the [SampleAssets website](https://github.khronos.org/glTF-Assets/).  I think that it is not necessary to be highly formal about screen shot size. It is most important that it correctly shows the asset.

DRx3D:
These comments were determined by Claude to be raised in various reviews but not resolved in the code. It is up to the author to make the determination if these are critical enough to resolve here or pushed off to a future release.

- main.ts arg parsing — lexaknyazev flagged that an unrecognized --foo-shaped argument silently falls through and is treated as a model name ("not sure if those cases need to be forbidden"). Author's reply only addressed the adjacent "short options" question (which was fixed — short forms were dropped). The dash-prefix/typo case itself was never revisited; parseCommandLine still does this today, so a typo'd flag in CI produces a confusing "model not found" instead of a clear usage error.
- README.body.md requirement — both reviewers explicitly +1'd (a) removing the `createReadme` opt-out flag and (b) requiring README.body.md instead of silently defaulting. (a) shipped — no createReadme flag remains anywhere. (b) did not: `Model.createReadmeMarkdown()` still silently inserts a bare ## Screenshot block when the file is missing, no validation error exists for it, and CONTRIBUTING.md's own text still says "this file is optional" — directly contradicting the agreed decision.
- Double-loop in ModelMetadata.process() (scans modelIssues once for anyHasErrors, loops again to print) — reviewer asked for either a single-loop simplification or a comment explaining the intentional two-pass behavior. Neither happened.
- The `warnings` mechanism itself — author's final comment: "strictly speaking, the warnings are now unnecessary. If we decide that there will never be Warnings, they could be removed completely." Nobody decided, and it's dead code today: `Issues.warnings` is still declared, `printIssues` still prints it, but nothing in the entire diff ever pushes into `issues.warnings` — it's permanently [].
- Variant file casing — lexaknyazev's final, explicit position: "Let's treat that as an error" for wrong-case extensions specifically (distinct from "file missing"). What shipped instead (`findVariantFileName`) does a case-sensitive `endsWith()` and, on no match, reports the generic `Did not find a ".glb" file for "..." variant` — indistinguishable from a genuinely missing file, losing the diagnostic value that was requested.
Duplicate glb/gltf branches in Model.ts — both agreed dedup "could be done later." Given how much this file has been rewritten since, worth confirming it's still tracked rather than forgotten (low priority).

DRx3D:
Items not raised in reviews, but found by Claude:

- No` package-lock.json`: `.gitignore` now excludes it, CI runs `npm install` (not `npm ci`), and `package.json` uses caret ranges (^25.8.0, ^10.9.2, ^6.0.3). Every CI run re-resolves the full dependency tree with no pinning — notable since `RebuildAllAssets.yml` uses this pipeline to write commits back to `main`.
- No automated tests: ~3,500 lines of new validation/generation logic ship with no test script and no test framework. Correctness rests entirely on `npm run check/update` against live data plus manual diff review.
- Hardcoded 2026 in generated README.md/REUSE.toml copyright strings — a deliberate, discussed tradeoff (avoids touching every file each year), but will silently go stale in 2027+ unless someone remembers to bump it by hand.
- Open design question, no resolution either way: whether the auto-populated `licenseUrl/spdx/icon` fields on legal entries should even be kept in `metadata.json`, since the PR itself notes removing icon could break github.khronos.org's `model viewer`, which reads it directly.

CONTRIBUTING.md
javagl: Based on another comment, the 'properly formatted' is now described as having a width of 150...500 pixels and a not-too-far-from-1.0 aspect ratio...

CONTRIBUTING.md
javagl: Based on another comment, the contributing guide was updated to make this file _required_, and I've added a check for its presence. (It was not present for some of the models, so I added it manually for those...)

javagl:
Thanks for the review.

---

The "obsolete" code parts mentioned in [the first comment](https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280#issuecomment-5160606732) are still on the radar. The currently intended workflow is roughly like this:

- Run the tool _locally_, including the `updateLegacyLicenseStructure` call
- This will re-generate the `metadata.json` files in their canonical form (all with proper license IDs and such...)
- Commit these new `metadata.json` files
  - It's not clear whether this will confuse the PHP version of the 'utils', but I don't think so, and if so, it shouldn't mattter
- Immediately after that: Remove the `updateLegacyLicenseStructure` function
- Commit the new 'utils' source code
  
---

> Screenshots have never been formally defined. [...]

I agree that we don't have to be overly formal. But this also came up in an [inlined comment by Alexey](https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280#discussion_r3298955924), because the contributing guide asks for a "properly formatted screenshot", without saying what that means.

I think that 150px is a bit too small, given today's resolutions. On the other hand, it may be more than enough for something like the `Triangle` model. So I'd say ~"something between 150px and 500px" or so. And unless we **explicitly** and **strictly** require and check (!) the exact size, the tables may look a bit odd. 

My response to that comment already links to https://github.com/KhronosGroup/glTF-Sample-Assets/issues/49, which shows that generating the tables with fixed image widths may look less jagged. 

So I updated the wording in the contributing guide and added that `width="250px"` to the table.

---

About the Claude comments - these are mainly picking up what Alexey already pointed out:

> main.ts arg parsing — lexaknyazev flagged that an unrecognized --foo-shaped argument silently falls through and is treated as a model name 

The original PHP did offer the functionality to run the tools on _individual_ models or "all models in the directory". Some options:

- We could establish a dedicated command for single models, like `main.ts --singleModel AnimatedTriangle` or so
- We could remove the function for processing single models. There's hardly any use-case for that, but when something's wrong with the `metadata.json` of a single model, it may be useful for focussed debugging
- We could accept that running `main.ts -check` treats `-check` as the model name

The error message in that case would be 
```
Model '-check' caused  1 errors and 0 warnings:
 E-1: Model metadata file not found: ./Models/-check/metadata.json
```
which IMHO is a clear enough hint at what happened to not require further action

> README.body.md requirement

Yes, this was agreed upon. However, there currently _are_ several models that do _not_ have a `README.body.md`, so it can not (trivially) be made a strict requirement...

**...unless** we add these files to the respective models. 
I just checked: It's really only a few models. Specifically:
```
Model 'MetalRoughSpheresNoTextures' caused  1 errors and 0 warnings:
 E-1: Model README body file not found: ./Models/MetalRoughSpheresNoTextures/README.body.md
Model 'RiggedFigure' caused  1 errors and 0 warnings:
 E-1: Model README body file not found: ./Models/RiggedFigure/README.body.md
Model 'RiggedSimple' caused  1 errors and 0 warnings:
 E-1: Model README body file not found: ./Models/RiggedSimple/README.body.md
Model 'SciFiHelmet' caused  1 errors and 0 warnings:
 E-1: Model README body file not found: ./Models/SciFiHelmet/README.body.md
Model 'TwoSidedPlane' caused  1 errors and 0 warnings:
 E-1: Model README body file not found: ./Models/TwoSidedPlane/README.body.md
Model 'VirtualCity' caused  1 errors and 0 warnings:
 E-1: Model README body file not found: ./Models/VirtualCity/README.body.md
```

So I updated the contributing guide, added these files accordingly, and added a check in the validation pass that checks for the presence of this file.

> Double-loop in ModelMetadata.process()  ...
> The warnings mechanism itself...

Yes, the double-loop was required for mimicking the behavior of the original PHP. And currently, we don't use the concept of 'warning', so that _could_ indeed be omitted. The PHP version _did_ have 'warnings'. If you have a strong preference, let me know.

> Variant file casing 

Yeah. The concept of variants is not defined strictly enough. The original set of variants (`glTF`, `glTF-Embedded`, `glTF-Binary`) did have clear file extensions associated with them, and are now hard-wired as the `knownVariantFileNameExtensions`. But there has been some proliferation for the variants, and when we have a variant like `glTF-KTX-BasisU` it's not clear what the file extension should be there (`gltf` or `glb`). 

I added a hint to the error message mentioning that the file extension check is _case-sensitive_ - I think that's a reasonable resolution for now...

---

About the other Claude comments:

> No package-lock.json: .gitignore now excludes it, 

This PR did not change anything in this regard. There are pros and cons for having such a lockfile. I don't have a strong preference. If someone has, it can either be added here or discussed in a dedicated issue.

> No automated tests: ~3,500 lines of new validation/generation logic ship with no test script and no test framework.

Yes. Just like before. 
(We can set up a $$$ contract and I'll add diligent unit tests that check if `getGlbUrl` returns the URI-encoded result of `getGlbPath`. (Well, I'd rather let some AI do it, but the $$$ could still go to me...))

> Hardcoded 2026 in generated README.md/REUSE.toml copyright strings — a deliberate, discussed tradeoff (avoids touching every file each year), but will silently go stale in 2027+ unless someone remembers to bump it by hand.

To my understanding, this doesn't have to be updated - it (roughly) has to be ~"the first date when stuff was published". If someone sees this as an issue, I'd consider this as independent of this PR itself.

> Open design question, no resolution either way: whether the auto-populated licenseUrl/spdx/icon fields on legal entries should even be kept in metadata.json, since the PR itself notes removing icon could break github.khronos.org's model viewer, which reads it directly.

It's not clear what the recommendation could be here. These properties are there. They probably don't **have** to be there (because they could be derived automatically). But existing tools may already **expect** them to be there, so they cannot trivially be removed without risking to break existing toolchains.

javagl:
The last commits here have been...

- Remove the `updateLegacyLicenseStructure` function that brought all `metadata.json` files into their canonical form (as a one-time operation)
- Update all `metadata.json` files with their canonical form, as of https://github.com/javagl/glTF-Sample-Assets/pull/4 
- Add a REUSE compliance check to CI, as of https://github.com/javagl/glTF-Sample-Assets/pull/5 

I think that this PR could be merged now. If everything goes well, then this will trigger the [`RebuildAllAssets.yml` CI step](https://github.com/javagl/glTF-Sample-Assets/blob/f83ea16e9bd24567796a940e0d3c76b70b47b55b/.github/workflows/RebuildAllAssets.yml) (which is activated on `push` to `main`), which will update all auto-generated files in the repository:

- The `README.md` and `LICENSE.md` files of all models
- The `model-index.json` that lists all models and their variants
- The listing files, like `Models-showcase.md`
- The `REUSE.toml`

PR #282, opened 2026-05-24
Add REUSE compliance check to CI

javagl:
Builds on top of https://github.com/KhronosGroup/glTF-Sample-Assets/pull/280

This only adds the 'REUSE' compliance check to CI.

Details to be discussed here, depending on whether there's that sought-for ✔️ , or that disappointing ❌ ...


javagl:
Yeah, that seems to work (basically). I still have to check what happens when a newly added model causes the check to fail, but... why should we anticipate that? We're optimists, right? 😉

javagl:
This was a basic test. The actual PR (going into the `new-utils` branch) is now at https://github.com/javagl/glTF-Sample-Assets/pull/5

PR #294, opened 2026-08-28
Traffic cone update

emackey:
Updates to the traffic cone, to address https://github.com/KhronosGroup/glTF-Sample-Assets/pull/285#issuecomment-5452590642


emackey:
The new screenshots here are also from vk_gltf_renderer, so the README is unchanged.

lexaknyazev:
Could you please comment on why `Cone_Retro.png‎` was added and `ConeBaseMat_metallicRoughness.png` was altered?

emackey:
Apologies, I didn't list my changes.

In the original configuration, `ConeBaseMat_metallicRoughness.png` was used for both retroreflection and roughness (metalness is zeroed out via factor).  The black areas make the orange parts of the cone avoid retroreflection, but simultaneously make them mirror-smooth (zero roughness) which is not appropriate for a roadside cone.  So, I raised the roughness of the orange parts of the cone, and then needed a separate image to avoid adding retroreflection at the same time.  So `Cone_Retro.png` is just the original image, applied only to retroreflection.

Additionally, I had Blender recompute the normal vectors for the model, as there were unintentional polygon edges visible within the base of the cone.  I also added some polygons to close the interior of the cone. The cone has a hole at the top, and has some interior walls extending down from that hole, but the walls abruptly ended, and you could see where polygons were missing by carefully peeking in the top of the cone.  Perhaps a small diagram would help explain:

diagram showing missing area

I added polygons inside the cone to close up the non-manifold hole in the mesh, shown as "MISSING" in red above.  Then I turned off double-sided for the cone materials.

Also, I made the two cones share mesh data (same accessors), to save space.  Previously there had been duplicate copies of the cone's vertex data.

Finally, the ground below the cone was reduced in area and textured with a CC0 asphalt texture from PolyHaven, noted in the metadata. It was moved down by a tiny fraction of a meter, to avoid z-fighting with the base of the cone.

javagl:
I think that these things have a donut topology, i.e. they are actually just the "hull", and a cross-section would be ...

KhronosCone

(Yes, I'm a smart-🫏, and that may not be the main point of the model, but _might_ be relevant for the question about the material being two-sided...)

emackey:
Indeed, that would be the correct topology, else they would fill with rainwater.  I added the polygons where I did because it was quicker & easier to close a hole than to cut into the bottom surface.  Should I fix it?

lexaknyazev:
> Also, I made the two cones share mesh data (same accessors), to save space. Previously there had been duplicate copies of the cone's vertex data.

Could that be a side-effect of Blender import? FWIW, the asset version that's currently in the `main` branch does not have duplicate cone's vertex data.

---

> I added polygons inside the cone to close up the non-manifold hole in the mesh, shown as "MISSING" in red above.

> I think that these things have a donut topology

Real-world traffic cones usually have no internal "floor" so that they are stackable for more efficient storage.

---

A few technical nitpicks:

- **Image size reduction.** All new JPEG files could be losslessly optimized. To do that, one could run the `jpegtran` tool from the `mozjpeg` project with  `-copy none -optimize -progressive -restart 0` parameters on each file. This operation would reduce the asset size by around 100 kB. PNG file sizes could be safely reduced by running `oxipng --zopfli --opt=max --strip=all *.png` in the `glTF` directory; it would also remove all unnecessary metadata. Also it would make sense to make a force-push with image updates to not increase the repo size twice.
- **Semantics of `/asset/extras`.** While it's clear how `/asset/extras` are useful for direct Sketchfab exports, I think having them in this asset could be misleading because the original file has been modified multiple times (including geometry changes); official license & copyright information is available in the usual place anyway. Besides, is the `asset/extras/additionally` property recognized anywhere? If not, we should not increase the non-standard part of the asset.
- **Use of TABs.** If merged as is, this asset would be the 6th instance of glTF JSON formatted with tabs instead of spaces in this repo. I don't have any strong preference on this, just highlighting the current state of formatting choices.

javagl:
> else they would fill with rainwater

Right, and you couldn't _stack_ them!!!111 😨

Of course, no strong preference from my side. 

For these tests, I'd usually say "As simple as possible (and ... whatever works)". (Somehow related, I recently extended the PBR extension support in JglTF, and for that, created a test model that arbitrarily combines [nearly all PBR extensions](https://github.com/javagl/JglTF/pull/154#issuecomment-5158246102), and ... what the ... is even that?!). And the cone is suitable for illustrating **what** the extension does, with an **actual** realistic test case.

(Note: Here I saw the answer from Alexey - that stacking is important... 😁  )

I think that it can remain the way it is (and actually, wouldn't even have seen the `doubleSided:true` as an issue - I'm generally considering this as the default, and it's hard to imagine _actual_ drawbacks of that...)

javagl:
Small addendum:

>> Also, I made the two cones share mesh data (same accessors), to save space. Previously there had been duplicate copies of the cone's vertex data.
> 
> Could that be a side-effect of Blender import? FWIW, the asset version that's currently in the main branch does not have duplicate cone's vertex data.

If this _is_ caused by the Blender import, then we might want to have a look at this. Importing the same mesh data (i.e. same accessors) should _probably_ not duplicate them by default. 

I just tried it out (in Blender 5.2.0): The glTF export dialog has a `Data -> Mesh -> Shared Accessors` check box, but that doesn't seem to make a difference here. So it _might_ be that the duplication is already happening on _import_ (and there doesn't seem to be an obvious way to avoid it, but would like to stand corrected...). 
(Running a trivial glTF-Transform `dedup()` on the result will deduplicate the accessors, but ... on the level of the _meshes_ - likely, because the extension isn't supported...)

EDIT: (Sorry for the spam): **Re-**importing the dedup'ed model _will_ cause the shared accessors to be retained. Handwavingly: Apparently, the Blender _import_ does not consider references to objects to be identical on the level of _accessors_, but only on the level of _meshes_.

mklefrancois:
image
I gave it a try and I think it looks better with a texture transform (scale 5,5) on the asphalt, and using the default camera.

emackey:
OK, the topology has been changed as noted above.  Also the scale of the pavement was adjusted. The screenshots should probably be retaken, but I didn't get a chance yet.

 - `/assets/extras` - It's perfectly valid glTF.  Why add rules here that aren't in the spec?

 - JPEG optimization - Feel free to adjust these, either on this branch or a new one.

 - TABs vs Spaces - I didn't make a conscious choice here, maybe the tabs came out of Blender, I didn't check.
 
I'll restate my concern that if we make the rules for Sample-Assets too rigid (for example, always TABs or always spaces, always POSIX-compliant text files, and more), then we're training the glTF ecosystem on rules that aren't in the specification.  A little mess here is good!  Our glTF readers must tolerate this.

lexaknyazev:
> `/assets/extras` - It's perfectly valid glTF. Why add rules here that aren't in the spec?

I'm not proposing any broad rules here, just concerned with the `/asset/extras/additionally` property. Is it a part of any ad-hoc schema used elsewhere? If not, we should not invent that just for this asset.

javagl:
> A little mess here is good! Our glTF readers must tolerate this.

👍 for that _in general_, and knowing what this refers to(!). But JSON and its validity are pretty narrow, and tab-vs-spaces are pretty unimportant (I cannot imagine that there is a JSON parser that accepts 2 spaces as indentation, but not 3....), so we _could_ consider some trivial auto-formatting of the glTF json, to avoid having completely unformatted glTF files in the repo...

mklefrancois:
Could we have a camera added to the scene set at position {0.0, 0.4, 1.6}, eye: {0.0, 0.2, 0.0} ?
For renderers that uses the scene camera, it would place the rendering position at the perfect place.


``` 
{
      "translation": [
        0.0,
        0.4,
        1.6
      ],
      "rotation": [
        -0.06213743984699249,
        -0.0,
        -0.0,
        0.9980676174163818
      ],
      "camera": 0,
      "name": "Camera"
    },


  "cameras": [
    {
      "type": "perspective",
      "name": "Example View",
      "perspective": {
        "zfar": 200.0,
        "znear": 0.10000000149011612,
        "aspectRatio": 1.7777777777777777,
        "yfov": 0.7853981633974483
      }
    }
``` 

image

emackey:
I moved the "Example View" camera to just in front of the bulb, which should be the "perfect" place for this.  It means you won't see the actual bulb icon from that camera's point of view, but you are 100% aligned with the retroreflection.

Also the bulb icon's material is unlit now.

emackey:
Per discussion at the PBR TSG meeting, I've collapsed the `asset.extras` block into a single, long `asset.copyright` string.

I decided to keep the comment on the additional sampler.  This really serves the same purpose as a comment in a piece of source code.  If I were editing source code and I turned off mip-mapping for a special case, it's very likely I would place a comment in the code, briefly explaining my reasoning.  Likewise, any human or AI system reading through the raw JSON here looking to gain more insight into the design might be led to wonder why there is an extra sampler block here, and what purpose it might serve.  I've placed a comment on the sampler block.  I don't want to move that to the README, I think it works best where it is in the code.

lexaknyazev:
> I think it works best where it is in the code.

Then only readers of the raw JSON would know about this detail. Given that `extras.comment` is not a standard feature (maybe it should become one eventually), many tools will ignore it and thus will never present to users. An additional comment in the README won't hurt, in my opinion.

emackey:
> many tools will ignore it and thus will never present to users.

That's exactly my analogy.  Most compilers ignore comments and don't embed them in executables.  One would not suggest to remove all comments from a large open-source project and instead publish those comments in the project documentation, where they would be readable by end users instead of folks copying and/or modifying the code. It's a different target audience.

emackey:
I removed the last remaining comment from the file.  Can this be merged/squashed before the meeting?  It contains important improvements to the sample that should be included in this repo when implementors begin adopting retroreflection.

